PR #3, opened 2024-11-21
d1: attempting to add d1 bindings

harryhcs:
trying to get d1 binding working locally and then getting some results from the db logged out
remote/production build seems to work fine


justinvdm:
@harryhcs @peterp Got it working with Prisma!

Code for using Prisma client in react component:

Screenshot 2024-11-25 at 17 38 22

Working in dev:

Screenshot 2024-11-25 at 17 38 11

Working in CF deployment:

Screenshot 2024-11-25 at 17 51 06

Will add context on the approach taken soon! Merging in the meantime to unblock.

PR #11, opened 2024-11-28
feat: Dev infra improvements

justinvdm:
* Avoid the need for separate React RSC vendor bundle (unlike react SSR, we can just use the main bundle)
* Don't block dev server startup on setup - it only needs to be done by the first request
* Don't block dev server and requests to it on types codegen (they arent used at runtime)
* Hide verbose output away - you can do `VERBOSE=1 pnpm dev` to see it.

https://www.loom.com/share/43b209884a524e91b4012de3616b336e?sid=b297fc02-d0f1-4ef5-bd1c-1e47d7da2e34


experiments/rsc/pnpm-lock.yaml
```diff
@@ -1,2797 +1,1376 @@
-lockfileVersion: '9.0'
+lockfileVersion: '6.0'
```
peterp, line 1: I guess we need to pin the version of pnpm somehow?

PR #22, opened 2024-12-10
Added persistence path for r2.

peterp:


justinvdm:
+1 if we still want to merge

PR #23, opened 2024-12-12
Support HMR

justinvdm:
## Context
* Adds Vite HMR support for both:
  * worker code run in miniflare/workerd
  * client-side code

## Demo
 https://app.claap.io/na-58/c-qkcr0NJtVm-gNq9Wh5FGL37

### Worker HMR
* When worker-related code changes, we broadcast to all clients that an RSC update is needed, the clients then ask for a new RSC payload
* **Why do it this way?** The clients (browser) are the ones with the context of what they need (e.g. what page they are currently on, so they're better positioned to ask the worker for what they want than having the server try to maintain this state

### Client-side HMR
* We just use `@vitejs/plugin-react` and the fast refresh capabilities it provides

### Under the hood
* Implements a vite plugin that uses [Environment API](https://vite.dev/guide/api-environment-runtimes) so that we can resolve+transform modules using vite dev server, but evaluate them in miniflare

## Attribution
* The vite plugin was basically copy-pasted from these sources (but grokked by rewriting line by line!):
  * https://github.com/hi-ogawa/vite-environment-examples/blob/81f30c9f9618878fd823a28d5970d6432d0d7002/packages/workerd/src/plugin.ts#L33
  * https://github.com/flarelabs-net/vite-plugin-cloudflare/blob/7c88a10cd1544ba920855709b7d28e3dbe9411f0/packages/vite-plugin-cloudflare/src/index.ts#L26

## What still needs to be done
* Adding "use client" or "use server" currently requires restarting dev server - I have a plan for how to avoid this
* We still re-run worker for "use client" code, we can avoid this
* Sometimes changing css classes in markup doesnt cause the change to reflect, I need to figure out why


experiments/rsc/package.json
```diff
@@ -39,16 +39,19 @@
...
     "dotenv": "^16.4.7",
     "es-module-lexer": "^1.5.4",
     "execa": "^9.5.1",
     "express": "^4.21.1",
     "fs-extra": "^11.2.0",
+    "import-meta-resolve": "^4.1.0",
     "kysely-codegen": "^0.17.0",
     "magic-string": "^0.30.14",
-    "miniflare": "^3.20241106.1",
+    "miniflare": "^3.20241205.0",
```
peterp, line 53: Curious about the downgrade?

peterp (approved):
❤️

peterp:
I'm going to merge this because I want to play with it!

experiments/rsc/package.json
justinvdm: @peterp its an upgrade, not a downgrade :D I think I needed it for types or exports they added related to worker options, and in the end didnt use it (yet).

PR #25, opened 2024-12-13
feat: Restart dev server on config changes

justinvdm:
Adds a vite plugin that will restart the dev server if any configs change (any js, ts or json file in root dir or in `scripts/`).

## Approach
* When a config file changes, close the dev server cleanly.
* Have the `pnpm dev` script in `package.json` re-run the script if it closed with a zero exit code => the user didn't Ctrl-C, and the script didn't exit with an error.
* Leverage Vite HMR browser client auto-retry capabilities to reload the page when the server is running again

## Notes
Ideally we could just use vite's `server.restart()`. I wasn't able to get this working - I got some obscure C error from workerd and didn't investigate further. That said, I think restarting the actual process is likely more robust, we can be more sure in this case that we're running on the latest changes.

### DIFF
--- a/experiments/rsc/package.json
+++ b/experiments/rsc/package.json
@@ -6,7 +6,7 @@
   "type": "module",
   "scripts": {
     "build": "npx tsx ./scripts/build.mts",
-    "dev": "NODE_ENV=development npx tsx ./scripts/runDevServer.mts",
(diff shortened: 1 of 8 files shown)


peterp:
Quality DX!

PR #32, opened 2025-01-08
Add basic user model

justinvdm:
* Adds basic `User` model with id and email
* Adds a new `pnpm migrate:new` script to automatically create new migration files based on:
  * (a) the current state of the local db
  * (b) changes to the prisma.schema file

### DIFF
--- a/experiments/billable/scripts/migrateNew.mts
+++ b/experiments/billable/scripts/migrateNew.mts
@@ -1,0 +1,33 @@
+import snakeCase from 'lodash/snakeCase';
+import { $ } from './lib/$.mjs';
+import { readdir } from 'fs/promises';
+import { resolve } from 'path'
+import { ROOT_DIR } from './lib/constants.mjs';
+
+const getNextMigrationNumber = async (): Promise<string> => {
+  const files = await readdir(resolve(ROOT_DIR, './migrations'));
+  const numbers = files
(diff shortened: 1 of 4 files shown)


experiments/billable/scripts/migrateNew.mts
```diff
@@ -0,0 +1,33 @@
...
+    .filter(num => !isNaN(num));
+
+  const lastNumber = Math.max(0, ...numbers);
+  return String(lastNumber + 1).padStart(4, '0');
+}
+
+export const migrateNew = async (name: string) => {
+  if (!name) {
+    console.log('Usage: pnpm migrate:new <migration-name>');
+    console.log('Example: pnpm migrate:new "Add user"');
```
justinvdm, line 20: We can now do this. Idea is:
1. we'd change the prisma.schema file
2. we'd run `pnpm migrate:new <migration name>`: it would generate a new migration file based on local db contents and changed prisma.schema file
3. we'd then run `pnpm migrate:dev` to apply the new migration

justinvdm: Ideally we could rely on Prisma's migrate tooling, but Prisma's recommendation is to use d1 migration tooling instead: https://www.prisma.io/docs/orm/overview/databases/cloudflare-d1#migration-workflows

peterp (approved):

PR #39, opened 2025-01-13
fix: Fix client side

justinvdm:
## Problem
I added a fix in #36 (https://github.com/redwoodjs/reloaded/pull/36/commits/44632d7973110fe4b72fd4037ff54e1cde31627f) that caused us to always resolve react to a custom build of react. This custom build caters for the fact that RSC and SSR are sharing a runtime.

However, I accidentally made the client side bundling also use this custom build of react, rather than normal react.

 ## Solution

I've added a plugin that allows us to alias react to the custom build of react only for the worker side, but leave it alone on the client side.

 ## Note
While fixing this, I discovered that the current way of having RSC and SSR share a single react build will not work for all things. For example, if there is a `use()` call in SSR-rendered code, we will end up with an issue because RSC is expecting its `use()`` and not the SSR `use()`.

I'll work on fixing this in a future PR.

### DIFF
--- a/experiments/billable/src/app/pages/invoiceDetail/FetchInvoice.tsx
+++ b/experiments/billable/src/app/pages/invoiceDetail/FetchInvoice.tsx
@@ -1,11 +1,8 @@
 "use server";
 
-import React from "react";
 import { getInvoice } from "../../services/invoices";
 
-export function FetchInvoice(props: { id: number, children: Function }) {
-
-  const invoice = React.use(getInvoice(props.id, 1));
+export async function FetchInvoice(props: { id: number, children: Function }) {
+  const invoice = await getInvoice(props.id, 1);
   return props.children(invoice)
 }
-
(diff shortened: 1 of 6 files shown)


experiments/billable/src/app/pages/invoiceDetail/FetchInvoice.tsx
```diff
@@ -1,11 +1,8 @@
 "use server";
 
-import React from "react";
 import { getInvoice } from "../../services/invoices";
 
-export function FetchInvoice(props: { id: number, children: Function }) {
-
-  const invoice = React.use(getInvoice(props.id, 1));
+export async function FetchInvoice(props: { id: number, children: Function }) {
+  const invoice = await getInvoice(props.id, 1);
```
justinvdm, line 6: While working on a fix for the original issue, I discovered that `use()` won't work either - for SSR renders, react gets the RSC version of use(), which it doesn't know how to use.

I'll work on a fix for this next (for use() and for more general usage of React APIs) so that SSR and RSC versions of react can co-exist better.

PR #40, opened 2025-01-13
fix: Avoid RSC react build when doing SSR

justinvdm:
Context coming soon!


experiments/billable/scripts/buildVendorBundles.mts
```diff
@@ -40,33 +40,11 @@ const configs = {
...
-          },
-        },
-        lib: {
-          entry: resolve(SRC_DIR, "react-rsc.ts"),
-          name: "react-rsc",
-          formats: ["es"],
-          fileName: "react-rsc",
-        },
-      },
-    }),
```
justinvdm, line 63: We no longer need that weird rsc bundle of react that I added back [here](https://github.com/redwoodjs/reloaded/pull/36/commits/44632d7973110fe4b72fd4037ff54e1cde31627f#diff-b6d4b9aa1e14ee1b4d3c5faa8433612357289ba0a2afe6d569b7ebd0f75343d6).

PR #70, opened 2025-01-29
Basic router.

peterp:


experiments/billable/src/worker.tsx
```diff
@@ -144,86 +89,110 @@ export default {
...
             });
 
-            return new Response(JSON.stringify({ key: r2ObjectKey }), {
-              status: 200,
+            console.log("performing login");
+
+            return performLogin(request, env, user.id);
+          }),
+
+          route("/invoices", InvoiceListPage),
```
justinvdm, line 134: @peterp so the handlers for routes are just components? That's awesome!!!

peterp: Yeah!

PR #80, opened 2025-02-05
feat: Extract shared scripts and configuration into @redwoodjs/reloaded

justinvdm:
### Video with context
_Coming soon!_

### Problem

Each Cloudflare application in our monorepo has its own scripts and configuration for Vite, React Server Components, routing, and authentication. This duplication makes maintenance difficult and slows down new application development. We also need a more structured approach as we prepare for external feedback on our framework.

### Solution

This PR introduces `@redwoodjs/reloaded`, a new package within the monorepo that consolidates common scripts and configuration. The `billable` experiment now uses this package, eliminating redundancy and ensuring consistency across projects.

### Changes

- **Extracted shared scripts and configurations** into `@redwoodjs/reloaded`.
- **Updated build commands**:
  - Previously: `pnpm dev`, `pnpm cf:deploy`, etc.
  - Now: `pnpm rw dev`, `pnpm rw deploy`, etc.
- **Simplified Vite configuration**:
  - Users now provide their own Vite config instead of relying on a wrapper.
  - Example vite config:
    ```tsx
    import { defineConfig } from "vite";
    import { redwood } from "@redwoodjs/reloaded/vite";

    export default defineConfig({
      plugins: [redwood()],
    });
    ```
- **Preserved existing functionality** while centralizing configuration.

### Notes
- The DX choices I made (e.g. `pnpm rw dev`) aren't dear to me - the goal was to make the code reusable. Lets iterate on this to get the DX what we want it to be :)

- This is the first step in modularizing the framework. Next steps:
  - Extract the library part (e.g., routing, authentication, React Server Component rendering, etc.)
  - Clean up and refine further

## Pointers for debugging

It is conceivable there will be broken things after this change.

If you need to debug things:

1. Use the `RW_DEV=1` env var to rebuild the ts code in @redwoodjs/reloaded when running `pnpm rw ...`

```sh
RW_DEV=1 pnpm rw dev
RW_DEV=1 pnpm rw build
```

2. To add/modify the commands available for `pnpm rw`, you can find them here: https://github.com/redwoodjs/reloaded/blob/737cc1a78701050dfcf0a602f2533bacefccca9c/reloaded/bin/rw.mjs#L18-L19

### Screenshots
Works in dev

Screenshot 2025-02-05 at 21 24 21

Works in production
Screenshot 2025-02-05 at 23 19 11

Projects now can have a standard `vite.config.mts` in their root and use redwood there as a plugin

Screenshot 2025-02-05 at 21 44 34


reloaded/package.json
```diff
@@ -0,0 +1,74 @@
...
+  },
+  "exports": {
+    "./vite": {
+      "default": "./dist/vite/index.mjs",
+      "types": "./dist/vite/index.d.mts"
+    }
+  },
+  "keywords": [],
+  "author": "",
+  "license": "UNLICENSED",
```
justinvdm, line 24: @peterp any thoughts on what license we want to use?

PR #83, opened 2025-02-10
feat: Refactor out runtime

justinvdm:
## Context
We're in the process of refactoring out a package for us to use across our experiments. #80 handled the scripts.

This PR handles refactoring out the main parts of the runtime code that runs in cloudflare workers into this reusable package.

## Changes
* Moves the majority of the non-app code out of `billable` to `@redwoodjs/reloaded/worker`
* Billable now defines its worker like this:
```ts
export default defineApp<ReturnType<typeof getContext>>({
  Head,  // provides the <title> and stylesheet <link> for the <head>
  getContext,  // { user }
  routes: [...]  // routes array, as is passed to defineRoutes()
})
```
* **note:** The idea was to just get _something_ in, we can totally still change the API entirely, was just the more straightforward way to abstract it out for a first iteration
* Moves routing, RSC-related runtime logic, and many other things into `@redwoodjs/reloaded/worker`
* It does _not_ move:
  * auth and session stuff: maybe we should rather scaffold this stuff out for now?
  * `/src/client.tsx`: on my list
  * api for worker scripts like seed script: on my list

Screenshot 2025-02-10 at 02 13 19


experiments/billable/src/worker.tsx
```diff
@@ -1,155 +1,54 @@
...
+  ]),
+  ...prefix("/user", authRoutes),
+  ...prefix("/invoice", invoiceRoutes),
+]
+
+export default defineApp<ReturnType<typeof getContext>>({
+  Head,
+  getContext,
+  routes,
+})
```
justinvdm, line 54: @peterp This is how the worker entry point looks now.

Totally don't mind this changing entirely, was just more straightforward to do for a first iteration.

Idea is that this just returns cloudflare worker handlers (e.g. `fetch`), so that the user can wrap this to do their own things. For example, the user can do:

```ts
const app = defineApp(...)

export default {
  fetch(request, env) {
    if (/* condition */) {
      return app.fetch(request, env)
   } else {
      return /* something else */
   }
  }
}
```

experiments/billable/src/env.ts
```diff
@@ -1,29 +0,0 @@
...
-}
-
-export function getEnv(): ParsedEnv {
-  if (!ENV) {
-    throw new Error("ENV is undefined.")
-  }
-  return ENV
-}
-
-
```
justinvdm, line 29: @peterp I removed this since it would take some thinking to figure out how we account for this ^ in the api. e.g. do we have zod schema for env. If we do, wondering why do we need it if cloudflare already gens the types for the env vars with its `Env` global type.

That ^, and the fact that the router provides the env as context, made me think that maybe its fine to do without it for now, but wanted to ask you about it.

PR #91, opened 2025-02-11
fix: Make React build compatible with CF vite plugin

justinvdm:
## Context
Vite dev server reloads as it discovers new deps during its dep optimization passes. When it does this, react and react-dom are re-instantiated. However, miniflare sandbox will have the old instances
around.

If react is re-instantiated after react-dom, then react-dom will get a different instance of the react internals object than
the one the new react is working off of.

* This is a problem for normal (SSR) React - new react module will not have its internal state set up correctly since
react-dom didn't interact with it
  => we need to hold onto the old instance, and export _it_ instead of the React instance, along with the old client internals
* On the other hand, for RSC react, it doesn't want to see that there was an inexisting renderer around before it was instantiated
  => we need to export the new instance of the server internals

 ## Why wasn't this a problem before?
I still need to confirm this, but this was what I discovered during debugging this:

I think the issue was still happening, but we were accidentally hiding it. The interesting thing about the problem is that it will only happen on the first request (when the dep optimization happens). After that, we would have be using the latest react internals so that further requests use it.

We happened to be making an initial request to the dev server before the user themselves made a request: https://github.com/redwoodjs/reloaded/blob/main/reloaded/src/vite/restartPlugin.mts#L58


reloaded/vendor/src/react.js
```diff
@@ -1,12 +1,29 @@
...
-__SERVER_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE.___RAR = 'RSC'
+// context(justinvdm, 2025-02-10): Vite dev server reloads as it discovers new deps during its dep optimization passes.
+// When it does this, react and react-dom are re-instantiated. However, miniflare sandbox will have the old instances around.
+// If react is re-instantiated after react-dom, then react-dom will get a different instance of the react internals object than
+// the one the new react is working off of.
+// * This is a problem for normal (SSR) React - new react module will not have its internal state set up correctly since
+// react-dom didn't interact with it
+//   => we need to hold onto the old instance, and export _it_ instead of the React instance, along with the old client internals
+// * On the other hand, for RSC react, it doesn't want to see that there was an inexisting renderer around before it was instantiated
+//   => we need to export the new instance of the server internals
```
justinvdm, line 14: This was the fix. The rest of the changes in this PR are tangential improvements I made during debugging

PR #99, opened 2025-02-12
chore: Use @cloudflare/vite-plugin's new WASM support

justinvdm:
Since 0.0.7, we no longer need our workarounds and patches for @cloudflare/vite-plugin in order to get Prisma WASM working.


pnpm-workspace.yaml
```diff
@@ -1,4 +1,3 @@
 packages:
   - 'reloaded'
   - 'experiments/billable'
-  - 'experiments/cutl'
```
justinvdm, line 4: @harryhcs I needed to remove this from here now that the vite plugin no longer should have that patch applied to it. Best explained over video, can you take a look when you have a moment? https://www.loom.com/share/109deb7365da476cafd73ab709da365b

reloaded/src/runtime/worker.tsx
```diff
@@ -45,25 +45,6 @@ export const defineApp = <Context,>(options: DefineAppOptions<Context>) => {
...
-        //
-        // ## Context:
-        // Vite sends an initial request to the worker when running the dev server,
-        // at which point the Prisma WASM is imported. Using the Prisma for the first time _after_ this initial request
-        // (e.g. if we only run the db for a request to /some/subpath) causes vite to not try import the WASM module
-        // at all, and ultimately the request ends up hanging indefinetely.
-        // * However, once the WASM has been imported, it is cached in some way that persists on the file system
-        // (from experimentation, it is not in node_modules/.vite). This means that if you were to subsequently
-        // change the code to _not_ have Prisma used after the initial request, the WASM will still be cached and
-        // the request will not hang. This makes this issue particularly hard to debug.
```
justinvdm, line 58: I'm testing out whether that hanging issue still happens.

PR #107, opened 2025-02-16
feat: minimal starter + support db-less apps

justinvdm:
## Adds a `starters/minimal` project
* Idea is that one can copy this project to use as a starting point for building an app
* The idea for the `minimal` starter is that it is a basic, bare-bones, essentials only skeleton to build upon
* It has no db, no styles, just a hello world page
* Idea is to add more starter projects, in particular batteries included ones (e.g. one that uses prisma)

### `import { db, setupDb } from './db'` - now in app code
We previous had one `import { db } from '@redwoodjs/reloaded'`, and we internally called `setupDb` to set it up. This code now lives in the app's code itself - this way we're not binding the user to having a db, and which client to use for that db

### Add a new (optional) `setup` option to `defineApp`
In order to hook in setting up of the db at the right place internally, we now need to support a `setup` option:

```ts
const setup = (env: Env) => {
  // the user can set up the db (or whatever else) here
}

const app = defineApp({
  setup,
  routers,
  getContext,
  Document,
})
```


experiments/yt-dos/src/app/pages/auth/routes.ts
```diff
@@ -1,54 +0,0 @@
...
-  route('/logout', function ({ request, env }) {
-    return new Response(null, {
-      status: 302,
-      headers: {
-        'Location': '/',
-        'Set-Cookie': `session_id=""; Path=/; HttpOnly; Secure; SameSite=Lax; Max-Age=0`,
-      }
-    });
-  }),
-]
```
justinvdm, line 54: @harryhcs I removed the auth stuff from the yt-dos project - I _think_ it wasn't being used, and I needed to remove it to fix it after I made db's optional with `@redwoodjs/reloaded`. I hope that makes sense!

PR #108, opened 2025-02-17
feat: Use router/middleware API for defining app

justinvdm:
# [video with context](https://www.loom.com/share/07cbb1cfa79d491a815ec85e8f73b032?sid=3b4cf6e9-d075-4a7c-9a03-9b40b4d48a45)

# migrating from the previous `defineApp` API

Previously, you would define an app like this:

```ts
import { defineApp } from '@redwoodjs/reloaded/worker';
import { index } from '@redwoodjs/reloaded/router';
import { Document } from 'src/Document';
import { Home } from 'src/pages/Home';

export const getContext = async (
  _request: Request,
  _env: Env,
) => {
  return {};
};


const routes = [
  index([
    Home,
  ]),
]

export default defineApp<ReturnType<typeof getContext>>({
  Document,
  getContext,
  routes,
})
```

It now looks like this:

```ts
import { defineApp } from '@redwoodjs/reloaded/worker';
import { index, layout } from '@redwoodjs/reloaded/router';
import { Document } from 'src/Document';
import { Home } from 'src/pages/Home';

type Context = {
}

export default defineApp<Context>([
  // >>> Replaces `getContext()`
  ({ ctx }) => {
    // >>> You can do side effects (e.g, setup like `setupDb()`) here
    // >>> You can set your context here
    ctx.foo = 23
  },
  // >>> This is how you say with Document to use now
  layout(Document, [
    // >>> these are your `routes`
    index([
      Home,
    ]),
  ]),
])
```

Here's an example that sets up the db and adds things to the context:

```ts
import { defineApp } from '@redwoodjs/reloaded/worker';
import { index, layout } from '@redwoodjs/reloaded/router';
import { Document } from 'src/Document';
import { Home } from 'src/pages/Home';
import { setupDb } from './db';

type Context = {
  user: Awaited<ReturnType<typeof getUser>>;
}

const getUser = async (request, env) => {
  // ...
}

export default defineApp<Context>([
  async ({ ctx, env, request }) => {
    setupDb(env)
    ctx.user = await getUser(request, env)
  },
  layout(Document, [
    index([
      Home,
    ]),
  ]),
])
```


experiments/billable/src/worker.tsx
```diff
@@ -26,40 +30,33 @@ export const getContext = async (
...
+              headers: { Location: link('/invoice/list') },
+            });
+          }
+        },
+        HomePage,
+    ]),
+    prefix("/user", authRoutes),
+    prefix("/invoice", invoiceRoutes),
+  ])
+])
```
justinvdm, line 59: This is the new API

peterp: How insanely good is this!

PR #123, opened 2025-02-25
feat: Passkey auth start

justinvdm:
Demo:
https://www.loom.com/share/a039d704671a445190d2063e94e044e0

More context to come soon!


sdk/src/runtime/lib/router.ts
```diff
@@ -167,27 +171,25 @@ export function defineRoutes<TContext = Record<string, any>>(
...
-            request,
-            params,
-            ctx,
-            env,
-            rw,
-          }) as Promise<Response>);
+          const r = await (h(routeContext) as Promise<Response>);
+          if (r instanceof Response) {
+            return r;
+          }
```
justinvdm, line 192: @peterp I needed to change this, otherwise middleware responses would short-circuit the chain, e.g. `index([() => { /* we would stop here */ }, Home])`

sdk/src/runtime/lib/router.ts
```diff
@@ -167,27 +171,25 @@ export function defineRoutes<TContext = Record<string, any>>(
...
+      const handlers = Array.isArray(handler) ? handler : [handler];
       for (const h of handlers) {
         if (isRouteComponent(h)) {
-          const actionResult = await rw.handleAction(ctx);
+          const actionResult = await rw.handleAction(routeContext);
           const serializedEnv = serializeEnv(env);
           const props = { params, env: serializedEnv, ctx };
           return await rw.renderPage({
-            Page: handler as React.FC<Record<string, any>>,
+            Page: h as React.FC<Record<string, any>>,
```
justinvdm, line 183: Another change I needed to make to get router working again - we're iterating over the handlers so I think we wanted the handler (`h`) itself not the array of handlers here.

sdk/src/runtime/lib/router.ts
peterp: 🆒 🆒 🆒 

This probably makes more sense anyway... Otherwise we try and send things that are not valid over the wire.

sdk/src/runtime/lib/router.ts
peterp: Ah fuck. That was my bad. Sorry.

PR #203, opened 2025-03-31
remove old experiments

harryhcs:
Remove my experiments from RedwoodSDK


peterp:
@harryhcs Are you going to host them somewhere else?

harryhcs:
> @harryhcs Are you going to host them somewhere else?

@peterp yess will host them on my own account, and also update to SDK, but not important now, I have them locally and running on CF

PR #255, opened 2025-04-07
Replace "pnpm run" commands.

### LINKED ISSUE #246: npm run dev requires pnpm
Was following the [quickstart](https://docs.rwsdk.com/getting-started/quick-start/) and chose the `npm` code option to `npm install` and `npm run dev`. The Vite server didn't run correctly as it still requires `pnpm` to be installed on my machine.

### LINKED ISSUE #248: `prisma generate` not running when applying migration
When applying a migration it appears that the Prisma client is generated, however it does not seem to be working.

peterp:
Fixes #246
Fixes #248


starters/standard/package.json
```diff
@@ -42,7 +41,8 @@
     "@types/react-dom": "^18.3.1",
     "prisma": "^6.4.1",
     "vite": "^6.1.0",
-    "wrangler": "^4.4.0"
+    "wrangler": "^4.4.0",
+    "typescript": "5.5.4"
```
peterp, line 45: @justinvdm We should probably remove "typescript" as a dependency in `sdk` and add it to the starters.

Issue #283, opened 2025-04-09
starters:  Older Durable Objects `migrations` wrangler config used?


dthyresson:
When I was trying out the Cloudflare Agents SDK I encountered an error with DO migrations.

I had to change `new_classes` to `new_sqlite_classes`

```json
 // Migrations configuration
  "migrations": [
    {
      "tag": "v1",
      "new_sqlite_classes": [ <-----
        "SessionDurableObject",
        "MyAgent"
      ]
    }
  ],
```

I think this is the new config.

<img width="754" alt="Image" src="https://github.com/user-attachments/assets/ea5a4d53-a3e9-46ba-be00-7a6b64fb84e3" />

See: https://developers.cloudflare.com/workers/wrangler/configuration/#migrations

Need to change starters where used like in standard for auth

justinvdm:
👍 this should be solved when we solve #274

justinvdm:
Done!

PR #289, opened 2025-04-09
chore: Migrate to sqlite-backed DOs

### LINKED ISSUE #274: Migrate DO storage to SQLite backend in starters and examples
Cloudflare now recommends using the SQLite backend for all new Durable Object classes ([GA announcement](https://developers.cloudflare.com/changelog/2025-04-07-sqlite-in-durable-objects-ga/)). Both our starters and examples currently use the older KV backend.

justinvdm:
Fixes #274 #283

### DIFF
--- a/starters/standard/wrangler.jsonc
+++ b/starters/standard/wrangler.jsonc
@@ -41,7 +41,7 @@
   "migrations": [
     {
       "tag": "v1",
-      "new_classes": ["SessionDurableObject"]
+      "new_sqlite_classes": ["SessionDurableObject"]
     }
   ],
 


starters/standard/wrangler.jsonc
```diff
@@ -41,7 +41,7 @@
   "migrations": [
     {
       "tag": "v1",
-      "new_classes": ["SessionDurableObject"]
+      "new_sqlite_classes": ["SessionDurableObject"]
```
justinvdm, line 44: From https://blog.cloudflare.com/sqlite-in-durable-objects/:

> SQLite-backed objects also support the existing [key/value-based storage API](https://developers.cloudflare.com/durable-objects/api/transactional-storage-api/): KV data is stored into a hidden table in the SQLite database. So, existing applications built on DOs will work when deployed using SQLite-backed objects.

From what I understand, the KV storage API is still supported - I think this makes sense - sometimes you don't want the structure, and just want keys and values.

In contrast, the KV _backend_ is superseded by the sqlite backend. This changes us to using the sqlite backend. So it seems its all we need to do, and the we can continue to use the KV API.

PR #301, opened 2025-04-13
(docs): updated code blocks for consistency

pantheredeye:
1. Added README style notes for code blocks with commands. 
2. Updated code blocks with commands to be more consistent. 

### Readme update

I wasn't really happy with the way this turned out, but I added it anyway. When previewed, the code blocks render as markdown, when I was really trying to show them as text. But, if someone looks at them in an editor, then they will see it as intended. 

### Doc update examples

Added withOutput to commands showing output:

image

Removed numbers from command blocks:

image

Added 'terminal' styling where missing:
image

Note: with all three style updates, i did not introduce any new formatting options. These styles already existed, they were simply applied inconsistently. For instance, I got the style for the command with output from the storage setup page.


justinvdm:
Thank you @pantheredeye!

Issue #284, opened 2025-04-09
Prefix WebAuthn environment variables for clarity


justinvdm:
The current APP_NAME variable in wrangler.jsonc is unclear in its purpose, especially when used with WebAuthn. To improve clarity, we should prefix WebAuthn-related environment variables with `WEBAUTHN_`.

justinvdm:
Solved in #304

PR #304, opened 2025-04-14
feat: Avoid manual deploy setup + env hoop jumping

### LINKED ISSUE #302: Quick Start last step yields an error
Following the [Quick Start](https://docs.rwsdk.com/getting-started/quick-start/) page from top to bottom (with the last command being `pnpm release` returns

```
✘ [ERROR] You must use a real database in the database_id configuration. You can find your databases using 'wrangler d1 list', or read how to develop locally with D1 here: https://developers.cloudflare.com/d1/configuration/local-development
```
In addition to this, the `wrangler.jsonc` file produced by the quick start script doesn't match what's on the Quick Start page. It has 3 additional _\_\_change_me\_\__ fields where as the Quick Start page only mentions to change the _name_ field (which isn't even the app name).

justinvdm:
## Context
A significant friction point in our DX is around env vars and deployment setup. Ideally it should be just as simple to do first deploy without any input, as it is do the same in dev. This PR solves this as follows.

## `npm run release` now has no manual steps
Before deploying, we now run a script that works as follows:
* if it finds `env.DB` usage in `src/`, it creates a d1 db (unless one has already been created)
* if it finds `WEBAUTHN` usage in `src/`, it creates the necessary auth env vars and sets them with `wrangler secret`

## Turnstile is now opt-in
The problem with turnstile is that there is unfortunately a manual step in CF UI. For this reason we leave Turnstile out, and recommend in the docs how to opt into it, and why you would want to

## No more need for `.env` in dev for auth
We now have defaults for the auth vars in dev, so you don't need an `.env` file unless you want to define your own APP-specific env vars.

## Links
* Discussion showing the current pain points: https://discord.com/channels/679514959968993311/1307974274145062912/1360728911071150202

Fixes #302, #284, #309


peterp:
This looks great @justinvdm!

I guess the one thing we have to alter is the `__change_me__` stuff in the wrangler config, which seems fine; but perhaps when we get a little more advanced we can just name that the same as the project name.

justinvdm:
> This looks great @justinvdm!
> 
> I guess the one thing we have to alter is the `__change_me__` stuff in the wrangler config, which seems fine; but perhaps when we get a little more advanced we can just name that the same as the project name.

@peterp as it happens, the PR does this too :D https://github.com/redwoodjs/sdk/blob/178211401485032438674eece09207b149d77d06/sdk/src/scripts/dev-init.mts#L17-L22

Just forgot to mention that, but we're thinking the same :)

PR #321, opened 2025-04-15
Ensure that "use client" is not accidentally transformed.

### LINKED ISSUE #269: Prevent accidental transform when 'use client' is not a directive
When a file contains the string 'use client' but not as a directive (e.g. inside a comment, variable, or function body), the transform mistakenly treats it as a directive. This causes the file to be transformed as a client component unexpectedly.

peterp:
The only thing we have to keep in mind is that this has to be the top of the file, if they add a comment and then `"use client";` it'll break.
Closes #269

### DIFF
--- a/sdk/src/vite/useClientPlugin.mts
+++ b/sdk/src/vite/useClientPlugin.mts
@@ -33,7 +33,7 @@ function isJsxFunction(text: string): boolean {
 export async function transformUseClientCode(
   code: string,
   relativeId: string,
-  isWorkerEnvironment: boolean,
+  isWorkerEnvironment: boolean
 ): Promise<TransformResult | undefined> {
   if (!isWorkerEnvironment) {
     return;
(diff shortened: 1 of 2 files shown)


sdk/src/vite/useClientPlugin.mts
```diff
@@ -288,12 +290,19 @@ export const useClientPlugin = (): Plugin => ({
       return;
     }
 
-    if (code.includes('"use client"') || code.includes("'use client'")) {
-      const relativeId = `/${relative(this.environment.getTopLevelConfig().root, id)}`;
+    const cleanCode = code.trimStart();
+    if (
+      cleanCode.startsWith('"use client"') ||
+      cleanCode.startsWith("'use client'")
+    ) {
```
peterp, line 297: This is all that I changed, I'm not sure why my prettier rules differ; let me see if I can fix this.

sdk/src/vite/useClientPlugin.test.mts
```diff
@@ -808,4 +808,13 @@ export function Chat() {
...
   });
+
+  it("Does not transform when 'use client' is not directive", async () => {
+    expect(await transform(`const message = "use client";`))
+      .toMatchInlineSnapshot(`
+      "import { registerClientReference } from "@redwoodjs/sdk/worker";
+
+      const message = "use client";"
+    `);
+  });
```
peterp, line 819: Added a test.

sdk/src/vite/useClientPlugin.mts
justinvdm: Oh right! We just need to check the start!

justinvdm (approved):

PR #346, opened 2025-04-23
fix: `use client` transformation duplicate exports

justinvdm:
## Context
When transforming `use client` modules for worker environment
* we keep the original components code as the SSR form the components (and export under name `<Component>SSR`) - this is used for SSR rendering
* we transform the existing component exports to instead be registered client references (since the actual component code is only relevant to client environment)

## Problem
* When adding SSR exports, we'd accidentally rename existing grouped exports too (e.g. `export { Button }` -> `export { ButtonSSR }`)
* We'd then later remove all original component exports and add them back ourselves together with their SSR equivalents
* But we'd look at the _original_ names when finding the original exports to remove them - except they are now named `<Component>SSR`

## Solution
Follow a more robust implementation with several passes:
1. collect info
2. remove exports
3. transform
4. add back exports

### DIFF
--- a/sdk/src/vite/useClientPlugin.mts
+++ b/sdk/src/vite/useClientPlugin.mts
@@ -37,7 +37,7 @@ export async function transformUseClientCode(
     return;
   }
 
-  onst project = new Project({
+  const project = new Project({
     useInMemoryFileSystem: true,
     compilerOptions: {
       sourceMap: true,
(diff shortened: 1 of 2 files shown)


justinvdm:
Merged here, GH is hanging on checking ability to merge: https://github.com/redwoodjs/sdk/commit/71f59ff660bc605e5a9186fb33638ad2e5985f8b

PR #347, opened 2025-04-23
fix: Ensure db exists

justinvdm:
Currently we only ensure db exists if the name is currently __change_me__ and we are going to replace it - not if db already configured in wrangler but not yet exists.


justinvdm:
On second thought this is a bad idea - if db already in config but not in d1, that isnt something we can do anything about because we don't know if we should replace db in config or not - better to inform user instead.

PR #371, opened 2025-04-28
fix env usage in docs

hffmnn:
When playing around with the sdk I noticed some docs that reference the usage of `env` via the `RequestInfo`. Trying to use this code gives a TS error: `Property 'env' does not exist on type 'RequestInfo<any, DefaultAppContext>'`.

I replaced the usage with this with the import of the `env` from `cloudflare:workers`


peterp (approved):
Awesome, thank you.

PR #380, opened 2025-04-30
docs: fix queues consumer example in queues.mdx

excessivecoding:
small fix on the queue consumer example in the docs that seems to be outdated

[Cloudflare doc for reference](https://developers.cloudflare.com/queues/get-started/#4-create-your-consumer-worker)


peterp (approved):
Thank you!

PR #381, opened 2025-04-30
docs: Queues - Explain queue naming and receiving patterns

dthyresson:
I followed the codes, but Cloudflare/Wrangler enforcing naming rules, so the `QUEUE_NAME` example (while a placeholder) if a dev tries it they will get an error:

```
🌀 Creating queue 'MY_QUEUE_NAME'

✘ [ERROR] A request to the Cloudflare API (/accounts/xxxx/queues) failed.

  Queue name 'my_queue_name' is invalid: invalid queue name: MY_QUEUE_NAME. Must match
  ^[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?$. [code: 11003]
  
  If you think this is a bug, please open an issue at:
  https://github.com/cloudflare/workers-sdk/issues/new/choose
```

Queues names cannot contain capital letters or underscores.

Note: Bindings can and a good practice is to have them standout like `MY_QUEUE_NAME`

Next, I updates the receiving messages section to include two ways to know which q you're handling the message:

1. Look at the batch for the name of the queue (dedicated queue per message type_
2. Add some `type` in message mode for "shared" queues

1 is preferred, IMO.

### DIFF
--- a/docs/src/content/docs/core/queues.mdx
+++ b/docs/src/content/docs/core/queues.mdx
@@ -19,10 +19,10 @@ To handle these you'll use **background tasks**. Background tasks are managed by
 First thing you've got to do is create a queue and bind the queue producers and consumers to your worker.
 
 ```bash showLineNumbers=false
-npx wrangler queues create MY_QUEUE_NAME
+npx wrangler queues create my-queue-name
 ```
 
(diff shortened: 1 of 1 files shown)


justinvdm:
🙇 !

PR #383, opened 2025-04-30
docs: Environment Variables - Runtime Types

dthyresson:
When I add new envars, I often forget that I have to run `pnpm wrangler types`.

While the envar is loaded fine, the types are unknown red squiggles.

Added this to the envar docs to explain how to and why.

Question: Should the starters also have a package script to do this?


justinvdm:
Thanks @dthyresson!

> Question: Should the starters also have a package script to do this?

Good question. Indeed we do have it, though it also runs other type codegens like prisma generate (for standard starter):

https://github.com/redwoodjs/sdk/blob/0820c3a36775c0231d0542866916c7434f14d4b5/starters/standard/package.json#L24

Might be worth us putting that there too - or s/npx wrangler types/npm run generate in the docs. Not sure.

PR #385, opened 2025-04-30
docs: Storage - added section on naming

dthyresson:
I often forget the different naming conventions with CF: R2 is different than Queues etc.

This adds a section on what valid R2 bucket names can be.


justinvdm:
Thanks DT!

PR #389, opened 2025-05-01
minimal-starter: update import paths to match standard starter

excessivecoding:
small update to make the alias match between the minimal starter and the standard one.

the idea is to make the documentation work for both (especially for the [shadcn/ui](https://docs.rwsdk.com/guides/frontend/shadcn/) docs) that uses the alias heavily due to the nature of shadcn/ui


peterp (commented):
Perfect, thank you!

PR #390, opened 2025-05-01
docs: fix mention of prisma starter

excessivecoding:
replace the mention of prisma starter with the standard starter as it seems the prisma starter doesn't exist (yet?)


peterp (approved):
Thank you!

PR #391, opened 2025-05-01
minimal-starter: ignore .wrangler

excessivecoding:
ignore .wrangler directly even if the minimal starter doesn't need it by default


peterp:
Thank you!

PR #393, opened 2025-05-01
vibe-rules support with ./llms export

FutureExcited:
adding support for [vibe-rules](https://github.com/FutureExcited/vibe-rules), which allows an easy installation of rules for your AI IDE.

`vibe-rules install <cursor | windsurf | claude-code | cline | roo>` 

exports ./llms, compatible with vibe-rules. you can configure whether it should be always attached by default, or set specific globs.

Using existing config from standard template rules

shape_dxmzeigJIA11JaN7yy2E6 at 25-05-01 23 06 20

PR #404, opened 2025-05-06
chore: Ignore .cursor in standard

justinvdm:
With the rules in `.cursor` now being autogenerated by `vibe-rules` (since #393), I _think_ we want `.cursor` dir gitignored.


justinvdm:
Actually, looks like one checks these in, closing.

PR #397, opened 2025-05-02
Docs for Storybook

arimendelow:
Includes:
- [x] Setup Storybook and launch in browser
- [x] Add a new component/story
- [x] Mocking dependencies, such as Prisma

Out of scope:
- Testing in Storybook? @ahaywood have you ever done this? I have not — also possible It'll overlap with your Testing docs

Also added PackageMananager helper component per https://github.com/redwoodjs/sdk/pull/397#discussion_r2083258450

PR #424, opened 2025-05-12
Docs: Add and use starlight-package-managers 

arimendelow:
Add and use https://github.com/HiDeoo/starlight-package-managers.

Per https://github.com/redwoodjs/sdk/pull/397#discussion_r2083258450 — docs should offer all 3 package manager options, and we want to be consistent.

Two notes:
- This package isn't quite perfect, though it does really save on line count. And it can certainly be made perfect 😊
  - Doesn't support multiline commands: https://github.com/HiDeoo/starlight-package-managers/issues/28
  - With `run` commands, it keeps `run` in the command for `pnpm` and `yarn`, where they aren't necessary
- _**As a proof of concept, I've so far only done the Quick Start page**_. If y'all like this, let me know and I'll convert over the rest of the docs.


peterp (approved):

peterp:
Thanks @arimendelow !

PR #427, opened 2025-05-12
🚧  WIP: Docs for Auth

ahaywood:


docs/src/content/docs/guides/auth/username-password/1-flow.mdx
```diff
@@ -0,0 +1,161 @@
...
+- }
+```
+
+You'll notice we also created an `@@index` for the username and email fields. This will help us speed up our queries, since we'll need to look up our users by username and/or email.
+
+We also removed the `Credential` model and its relationship on the `User` model.
+
+Now, let's run our migration:
+
+```bash
```
arimendelow, line 157: nit: instead of doing raw bash statements, can you use the PackageManagers component to stay consistent?

peterp:
Gonna close this for now. Let me know if you want to open it again.

PR #440, opened 2025-05-14
Docs: Be more agnostic about which package manager is used

arimendelow:
Following #424, convert more package manager commands to use the PackageManagers helper to both make the docs more agnostic a bout which package manager is used, and simplify situations where all 3 were manually being expressed with tabs.

Note that anything multiline still needs to manually use tabs, and that such cases need to be updated to use the PackageManagers syncKey.


justinvdm:
> Note that anything multiline still needs to manually use tabs, and that such cases need to be updated to use the PackageManagers syncKey.

@arimendelow sorry, not following here, could you elaborate?

arimendelow:
@justinvdm, PackageManagers doesn't support multiline blocks. See conversation here: https://github.com/HiDeoo/starlight-package-managers/issues/28

PR #443, opened 2025-05-15
standard-starter: Initialise Prisma using env import

mj-meyer:
In March, the [Cloudflare team added support](https://developers.cloudflare.com/changelog/2025-03-17-importable-env/) for importing bindings like `env` directly.

This allows us to initialise `PrismaClient` at the module level, making the Prisma WASM immediately available when the worker starts. That prevents the first request from hanging (as noted by @justinvdm).

This change tripped me up while integrating Better-Auth, since I used this starter as the base.

### DIFF
--- a/starters/standard/src/db.ts
+++ b/starters/standard/src/db.ts
@@ -1,23 +1,7 @@
 import { PrismaClient } from "@prisma/client";
 import { PrismaD1 } from "@prisma/adapter-d1";
+import { env } from "cloudflare:workers";
 
-export let db: PrismaClient;
-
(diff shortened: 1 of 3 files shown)


peterp:
@justinvdm I wonder if that weird initialization bug with prisma + wasm is now fixed. It would be so good to get off the wasm binary!

justinvdm:
Thank you @mj-meyer!

yieme:
@mj-meyer, cool. Could you share your RedwoodSDK starter with Better-Auth?

PR #450, opened 2025-05-16
Fix https to http for localhost in full stack docs

timReynolds:
This is a small fix to the full stack tutorial which mistakenly uses https rather than http for localhost links. If you click the existing links you'll experience an error locally


peterp (approved):
Thank you!

PR #452, opened 2025-05-18
Allow comments and whitespace in file before "use client"; and "use server";

peterp:
Allows comments and whitespace at the start of a client component directive.

### DIFF
--- a/sdk/src/vite/useClientPlugin.mts
+++ b/sdk/src/vite/useClientPlugin.mts
@@ -35,12 +35,7 @@ export async function transformUseClientCode(
   code: string,
   relativeId: string,
 ): Promise<TransformResult | undefined> {
-  const cleanCode = code.trimStart();
-
-  if (
-    !cleanCode.startsWith('"use client"') &&
-    !cleanCode.startsWith("'use client'")
-  ) {
+  if (code.indexOf("use client") === -1) {
     return;
   }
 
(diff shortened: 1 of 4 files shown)


peterp:
@justinvdm It's odd that the realtime upgrades are failing on the smoke-tests. I'll investigate.

sdk/src/vite/useClientPlugin.mts
```diff
@@ -35,12 +35,7 @@ export async function transformUseClientCode(
   code: string,
   relativeId: string,
 ): Promise<TransformResult | undefined> {
-  const cleanCode = code.trimStart();
-
-  if (
-    !cleanCode.startsWith('"use client"') &&
-    !cleanCode.startsWith("'use client'")
-  ) {
+  if (code.indexOf("use client") === -1) {
```
justinvdm, line 38: @peterp the thing with checking anywhere in file is it could be anywhere - even in a comment, or not at top level scope. I kind of think it needs to like it was because of that. It has tripped me up before - I had "use client" in comment.

I guess we could get more sophisticated and check that it is at top level scope?

justinvdm: Ah, looks like you solved this here: https://github.com/redwoodjs/sdk/pull/453/files#diff-c73445294aa5713be53bfcb06742e815e35ef3c23472bc28fcc6634b806b93d9

I'll close this one then - looks like that PR solves it too. Let me know if I misunderstood and if we still need this PR

justinvdm:
Closing since #453 solves this too

PR #453, opened 2025-05-18
RSC directives, support default exports in server components.

### LINKED ISSUE #84: "use server" breaks if `export default` used
When using "use server", you're able to `export`, but not `export default`. Otherwise you'll see this when you `pnpm build`

```ts
src/app/pages/invoice/routes.ts (3:7): "default" is not exported by "src/app/pages/invoice/ListPage/InvoiceListPage.tsx", imported by "src/app/pages/invoice/routes.ts".
file: /Users/justin/rw/reloaded/experiments/billable/src/app/pages/invoice/routes.ts:3:7

1: import { db, index, route } from "@redwoodjs/reloaded/worker";
2: import InvoiceDetailPage from "./DetailPage/InvoiceDetailPage";
3: import InvoiceListPage from "./ListPage/InvoiceListPage";
```

This happens because of how we transform "use server" modules to wrap exports to register server-side actions. We need to either fix this, or if it makes sense to keep the restriction, catch it and provide understandable error message about what happened, how to fix it, and why it needs to be that way.

peterp:
This PR rewrites the way that we're matching "use server"; in the worker and client. It now supports default exports:

Depends on #452
```
"use server"

export default function sum() {
  return 1 +1
}
```

becomes:
```
"use server"

export function sum() {
  return 1 +1
}

export default sum
```

@justinvdm 

Tests:
- [x] Single line comments
- [x] Multi-line comments
- [x] Comment blocks
- [x] export default function sum() {}
- [x] export function sum() {}
- [x] export async function sum {}
- [x] export const sum = () => {}
- [x] export const sum = async () => {}

@justinvdm Can you think of any patterns that I'm missing?

### DIFF
--- a/sdk/src/vite/__snapshots__/useServerPlugin.test.mts.snap
+++ b/sdk/src/vite/__snapshots__/useServerPlugin.test.mts.snap
@@ -1,0 +1,146 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`useServerPlugin > TRANSFORMS > ARROW_FUNCTION_EXPORT_CODE > CLIENT 1`] = `
+"import { createServerReference } from "rwsdk/client";
+
+export let sum = createServerReference("/test.tsx", "sum");
+"
+`;
+
+exports[`useServerPlugin > TRANSFORMS > ARROW_FUNCTION_EXPORT_CODE > WORKER 1`] = `
+"import { registerServerReference } from "rwsdk/worker";
+
+export const sum = () => {
+  return 1 + 1
+}
+registerServerReference(sum, "/test.tsx", "sum")
+"
+`;
+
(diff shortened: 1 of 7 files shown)


sdk/src/vite/__snapshots__/useServerPlugin.test.mts.snap
```diff
@@ -0,0 +1,146 @@
...
+registerServerReference(sum, "/test.tsx", "sum")
+"
+`;
+
+exports[`useServerPlugin > TRANSFORMS > DEFAULT_EXPORT_CODE > CLIENT 1`] = `
+"import { createServerReference } from "rwsdk/client";
+
+export let sum = createServerReference("/test.tsx", "sum");
+"
+`;
```
justinvdm, line 84: I think this might be an issue, since the client will be expecting to import the default:

```
import sum from "./test.tsx";
...
```

PR #454, opened 2025-05-18
Fix URL pattern matching

### LINKED ISSUE #312: Route param names with :'s breaks runtime
From https://discord.com/channels/679514959968993311/1361442272289493002/1361484201051947029

>  Doing something like this starts breaking down the parameters to individual characters. I think in reverse order.
route("/pokemon/:name:test:big:cool", Pokemon),

## Implementation note
Can we use a library instead for param parsing in routes? I think express used some library for this.

peterp:
Closes #312

I tried to make this work with the `URLPattern` API, but wasn't successful. I'll take another look at it when I get more time.

### DIFF
--- a/sdk/src/runtime/lib/router.test.ts
+++ b/sdk/src/runtime/lib/router.test.ts
@@ -1,0 +1,82 @@
+import { describe, it, expect } from "vitest";
+
+import { matchPath } from "./router";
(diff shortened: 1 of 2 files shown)


justinvdm (approved):
🍰

PR #455, opened 2025-05-18
Add disable RSC flight option

### LINKED ISSUE #398: RFC: Allow user to disable adding RSC payload to flight data.
Right now we always append the RSC flight data at the end of the HTML page; we give users the ability to "control every byte of the wire," so we should allow them to disable that. My suggestion is that we add a flag to the render function `render(Document, [routes], { RSCflight: false })`?

I'm not entirely happy with the option, maybe another function would be helpful?

`renderStatic()`

Based off this conversation: https://x.com/asidorenko_/status/1917647721227239669

peterp:
Closes #398.

- [x] Documentation


justinvdm (approved):
💥

PR #458, opened 2025-05-19
Doc typos

standuprey:
`type="install"` gives `yarn install` instead of `yarn add`, not putting any type shows the correct default install command for each package manager


peterp:
Thank you!

PR #461, opened 2025-05-21
Ensure prisma client initialized after wasm engine initialized

justinvdm:
For prisma-client-js generator or cases where there are dynamic import to the prisma wasm modules, we need to make sure we are instantiating the prisma client later in the flow when the wasm would have been initialized.

Fixes regression introduced in #443 that appears when querying in a deployed app

```
ReferenceError: Cannot access 'UL' before initialization
```

### DIFF
--- a/starters/standard/src/db.ts
+++ b/starters/standard/src/db.ts
@@ -4,6 +4,10 @@ import { env } from "cloudflare:workers";
 
 export let db: PrismaClient;
 
+// context(justinvdm, 21-05-2025): For prisma-generator-js or cases where there
+// are dynamic import to the prisma wasm modules, we need to make sure we are
+// instantiating the prisma client later in the flow when the wasm would have
+// been initialized.
 export const setupDb = async () => {
   const db = new PrismaClient({
     adapter: new PrismaD1(env.DB),
--- a/starters/standard/src/db.ts
+++ b/starters/standard/src/db.ts
@@ -2,6 +2,13 @@ import { PrismaClient } from "@prisma/client";
 import { PrismaD1 } from "@prisma/adapter-d1";
 import { env } from "cloudflare:workers";
 
-export const db = new PrismaClient({
-  adapter: new PrismaD1(env.DB),
-});
+export let db: PrismaClient;
+
+export const setupDb = async () => {
+  const db = new PrismaClient({
+    adapter: new PrismaD1(env.DB),
+  });
+
+  // context(justinvdm, 21-05-2025): https://github.com/cloudflare/workers-sdk/pull/8283
+  await db.$queryRaw`SELECT 1`;
+};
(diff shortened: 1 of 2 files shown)


Checkm3out:
This change did not work for me but I found one that did.


import { PrismaClient } from "@prisma/client";
import { PrismaD1 } from "@prisma/adapter-d1";

export let db: PrismaClient;

export const createDbClient = (env: Env) =>
  new PrismaClient({ adapter: new PrismaD1(env.DB) });

export const setupDb = async (env: Env) => {
  db = createDbClient(env);
  // todo(justinvdm, 30 Jan 2025): Figure out how to avoid this.
  //
  // ## Context:
  // Vite sends an initial request to the worker when running the dev server,
  // at which point the Prisma WASM is imported. Using the Prisma for the first time _after_ this initial request
  // (e.g. if we only run the db for a request to /some/subpath) causes vite to not try import the WASM module
  // at all, and ultimately the request ends up hanging indefinetely.
  // * However, once the WASM has been imported, it is cached in some way that persists on the file system
  // (from experimentation, it is not in node_modules/.vite). This means that if you were to subsequently
  // change the code to _not_ have Prisma used after the initial request, the WASM will still be cached and
  // the request will not hang. This makes this issue particularly hard to debug.
  await db.$queryRaw`SELECT 1`;
};


then in your worker pass the env to the setupDB     await setupDb(env);


__
With the version in this pull request i was not able to access the DB in local but with the one in this comment i have it working

justinvdm:
> This change did not work for me but I found one that did.
> 
> import { PrismaClient } from "@prisma/client"; import { PrismaD1 } from "@prisma/adapter-d1";
> 
> export let db: PrismaClient;
> 
> export const createDbClient = (env: Env) => new PrismaClient({ adapter: new PrismaD1(env.DB) });
> 
> export const setupDb = async (env: Env) => { db = createDbClient(env); // todo(justinvdm, 30 Jan 2025): Figure out how to avoid this. // // ## Context: // Vite sends an initial request to the worker when running the dev server, // at which point the Prisma WASM is imported. Using the Prisma for the first time _after_ this initial request // (e.g. if we only run the db for a request to /some/subpath) causes vite to not try import the WASM module // at all, and ultimately the request ends up hanging indefinetely. // * However, once the WASM has been imported, it is cached in some way that persists on the file system // (from experimentation, it is not in node_modules/.vite). This means that if you were to subsequently // change the code to _not_ have Prisma used after the initial request, the WASM will still be cached and // the request will not hang. This makes this issue particularly hard to debug. await db.$queryRaw`SELECT 1`; };
> 
> then in your worker pass the env to the setupDB await setupDb(env);
> 
> __ With the version in this pull request i was not able to access the DB in local but with the one in this comment i have it working

Thanks @Checkm3out, I wondered if we needed to back all the way to this. I'll do that for safety 👍

justinvdm:
On second thought, we're now in a position to land #457, which should avoid the need for this hoop, so will got that route instead

PR #496, opened 2025-06-06
docs: update minimal README to use minimal template

jonotron:
Previous readme didn't include `-t minimal` which defaults to `standard`


peterp:
Thanks!

PR #497, opened 2025-06-07
docs: remove note

bb-at-easley:
# Removes Quick Start Note in Authentication Section

image

Removes the note about Quick Start using the minimal template from the _Authentication_ section. According to[ create-rwsdk](https://docs.rwsdk.com/reference/create-rwsdk/), the default command uses standard template.

> The [Quick Start Guide](https://docs.rwsdk.com/getting-started/quick-start) uses the minimal starter, you should use the standard starter to get you set up and ready for using Database, Queues, Sessions and Authentication.

## Thought: 
**Caveat:** check if the standard starter still uses: Database, Queues, Sessions and Authentication. 👀 

If no, modify the note or the doc vs remove altogther.


peterp:
Thanks!

PR #505, opened 2025-06-09
docs(tailwind) Remove vite env config

arimendelow:
Not only did I find adding `environments` with `ssr` to not be necessary, but including it on v0.0.88 breaks my app:
Screenshot 2025-06-08 at 20 14 43

The errors says that it'll ask for a reload, but the error doesn't go away even if you restart the dev server, and in any case it's working just fine without it 🤷🏼‍♂️

### DIFF
--- a/docs/src/content/docs/guides/frontend/tailwind.mdx
+++ b/docs/src/content/docs/guides/frontend/tailwind.mdx
@@ -20,27 +20,18 @@ Since the RedwoodSDK is based on React and Vite, we can work through the ["Using
     />
 
 2. Configure the Vite Plugin
-    ```ts ins="import tailwindcss from '@tailwindcss/vite'" ins="tailwindcss()," {3, 12} ins={7-9} title="vite.config.mts"
+    ```ts ins="import tailwindcss from '@tailwindcss/vite'" ins="tailwindcss()," title="vite.config.mts"
     import { defineConfig } from "vite";
     import tailwindcss from '@tailwindcss/vite'
     import { redwood } from "rwsdk/vite";
 
     export default defineConfig({
-      environments: {
-        ssr: {},
-      },
       plugins: [
         redwood(),
         tailwindcss(),
       ],
     });
     ```
-
-    <Aside type="note" title="Environment Configuration">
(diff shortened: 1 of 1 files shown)


justinvdm:
>  and in any case it's working just fine without it 🤷🏼‍♂️

@arimendelow have you tried a `pnpm build`? I was able to repro it again on 0.0.88 and `@tailwindcs/vite` 4.1.8 here: https://github.com/redwoodjs/sdk-experiments/pull/4

If you have a `import stylesUrl from "./styles.css?url"` + `@import "tailwindcss"` in `styles.css` + `tailwind()` in your vite config, then try `pnpm build`, that should be enough to repro the issue.

Please let me know if you repro it with that ^, and I'll close this PR.

That all said, this workaround shouldn't be needed on `0.1.0-alpha` builds, so we won't need this in docs much longer.

Regarding the error you're running into, does this happen even after clearing `node_modules/.vite` and rerunning `pnpm dev`? If so, would you be able to push up a repro of the issue? I've never seen it before tbh, and we've had quite a few projects using tailwind so far with rwsdk.

arimendelow:
Good callouts. Will try all this and get back to you!

ahaywood:
@arimendelow Did you get a chance to try this?

arimendelow:
@ahaywood noooo i've been distracted by two other issues 🙈 — server function I/O errors that i don't understand and my requestInfo's header coming back as empty and refusing to be set. i'm thinking it's maybe time to start fresh on 0.1.0 and copy my code over bit by bit (tbh there's not a lot) to hopefully understand more about why and when things are breaking. but i digress 😃 

in any case, we don't need the `env` config anymore, right?

peterp:
@ahaywood Can you test this and confirm that it's not required anymore?

peterp:
I can confirm that this is still required.

arimendelow:
Whaaaat I wonder why I didn't need it in my project. I suppose I need to move to 0.1.0 stable.

Thanks Peter!

ahaywood:
@peterp Thanks, Peter!

PR #513, opened 2025-06-11
Update shadcn.mdx

johndevor:
Typos


ahaywood:
@johndevor Thanks for catching!

PR #521, opened 2025-06-15
perf: Improve startup times

justinvdm:
[x] Avoid transforms for app code in optimizeDeps
[x] `rwsdk/client` include to avoid optimizeDeps reload
[x] Invalidate client and server module lookups when files added after initial render
[ ] Ensure we keep deps in module lookups even when there's an optimizeDeps cache hit
[ ] Replace ts-morph implementation with ast-grep+magic-string for directive transformations
[ ] Short-circuit jsx transforms earlier using heuristic


peterp:
This is really cool!

PR #526, opened 2025-06-17
feat: Support renderToStream()

### LINKED ISSUE #422: RFC: Allow returning elements from interuptors
## Problem
Currently one can only return Responses from interruptors. But sometimes you want to short-circuit the flow, but return an element. For example, returning a 404 element from an interruptor if something isn't found.

## Suggested Solution
For example, 404's:

```ts
export default defineApp([
  // ...
  async () => {
    const page = await db.page.findUnique();
    if (!page) {
      return <NotFound />;
    }
  },
  // ...
]);
```

## Open Questions
* For 404 case, how do we set the status code?

For example:

```ts
export default defineApp([
  // ...
  async ({ response }) => {
    const page = await db.page.findUnique();
    if (!page) {
      // response is an object with properties to be set on Response
      response.status = 404;
      return <NotFound />;
    }
  },
  // ...
]);
```

that ^ doesn't feel great, not sure what else we could do though.

Semantically, what we're wanting is: `return new Response(<NotFound />, { status: 404 })`. But the actual solution can't look that way - that'd be a non-standard Response then.

justinvdm:
### `renderToStream(element[, options]): Promise<ReadableStream>`

Takes in a React Server Component (can be a client component or server component), and returns a stream that decodes to html.

```tsx
const stream = await renderToStream(<NotFound />, { Document })

const response = new Response(stream, {
  status: 404,
});
```

#### Options
* `Document`: The [document](/core/routing/#documents) component to wrap around the React Server Component `element`. If not given, will return the rendered React Server Component without any wrapping.
* `injectRSCPayload = false`: Whether to inject the corresponding RSC payload for the React Server Component to use for client-side hydration
* `onError`: A callback function called with the relevant error as the only paramter if any errors happen during rendering

### `renderToString(element[, options]): Promise<string>`

Takes in a React Server Component (can be a client component or server component), and returns an html string.

```tsx
const html = await renderToString(<NotFound />, { Document })

const response = new Response(html, {
  status: 404,
});
```

#### Options
* `Document`: The [document](/core/routing/#documents) component to wrap around the React Server Component `element`. If not given, will return the rendered React Server Component without any wrapping.
* `injectRSCPayload = false`: Whether to inject the corresponding RSC payload for the React Server Component to use for client-side hydration

Fixes #422

### DIFF
--- a/docs/src/content/docs/core/react-server-components.mdx
+++ b/docs/src/content/docs/core/react-server-components.mdx
@@ -100,3 +100,46 @@ export default function AddTodo() {
 ### Context
 
 Context is a way to share data globally between server components on a per-request basis. The context is populated by middleware, and is available to all React Server Components Pages and Server Functions via the `ctx` prop or `requestInfo.ctx`.
+
+
+## Advanced usage
+
+### Manual rendering
+
+RedwoodSDK also provides a way to render your React Server Components imperatively with `renderToStream()` and `renderToString()`.
+
+To render your component tree to a [`ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream)
+
+
+### `renderToStream(element[, options]): Promise<ReadableStream>`
+
+Takes in a React Server Component (can be a client component or server component), and returns a stream that decodes to html.
+
+```tsx
+const stream = await renderToStream(<NotFound />, { Document })
+
+const response = new Response(stream, {
+  status: 404,
+});
+```
+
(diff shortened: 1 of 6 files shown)


justinvdm:
Merged manually because GH merge check acted up: https://github.com/redwoodjs/sdk/commit/afbfd20309b29b73833874713bf4361cda2de4f4

peterp:
This is an amazing feature! Nice!

PR #527, opened 2025-06-17
Expose hydrateRoot options.

peterp:
This exposes `hydrationOptions` from `hydrateRoot`. This is required to turn off errors that occur when HTML rendering is disable. 

```tsx
import { initClient } from 'rwsdk/client'

initClient({ hydrationOtions: {...} })
```

cc: @jldec

### DIFF
--- a/sdk/src/runtime/client.tsx
+++ b/sdk/src/runtime/client.tsx
@@ -1,5 +1,6 @@
 import { clientWebpackRequire } from "./imports/client";
 import { type CallServerCallback } from "react-server-dom-webpack/client.browser";
+import { type HydrationOptions } from "react-dom/client";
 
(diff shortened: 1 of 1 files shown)


peterp:
Based off of this PR: https://github.com/jldec/agents-chat/pull/25

peterp:
@justinvdm Open to a discussion about this, I don't know if there's a better way.

justinvdm:
@peterp sorry for the delay, just wrapping my head around this all now

justinvdm:
@peterp is the idea that we'd catch the error?

If so, I don't think it'd hurt to expose they `hydrateRoot` options, so will merge, but I also feel like we should be able to avoid hydration mismatches in the first place for this "empty shell" case. I'll investigate.

PR #530, opened 2025-06-18
Add client side navigation

peterp:
This adds the ability for us to initialize client-side navigation, also known as single-page apps.

Docs: https://pp-client-navigation.redwood-sdk-docs.pages.dev/guides/frontend/client-side-nav/

### DIFF
--- a/docs/src/content/docs/guides/frontend/client-side-nav.mdx
+++ b/docs/src/content/docs/guides/frontend/client-side-nav.mdx
@@ -1,0 +1,38 @@
+---
+title: Client Side Navigation (Single Page Apps)
+description: Implement client side navigation in your RedwoodSDK project
+---
+
+## What is Client Side Navigation?
+
+Client-side navigation is a technique that allows users to move between pages without a full-page reload. Instead of the browser reloading the entire HTML document, the JavaScript runtime intercepts navigation events (like link clicks), fetches the next page's content (usually as JavaScript modules or RSC payload), and updates the current view.
+
+This approach is commonly referred to as a Single Page App (SPA). In RedwoodSDK, you get SPA-like navigation with server-fetched React Server Components (RSC), so it's fast and dynamic, but still uses the server for rendering.
+
+```tsx title="src/client.tsx" ins="initClientNavigation()"
+import { initClient, initClientNavigation } from "rwsdk/client";
+
+initClient();
+initClientNavigation();
+```
+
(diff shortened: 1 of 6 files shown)


docs/src/content/docs/guides/frontend/client-side-nav.mdx
```diff
@@ -0,0 +1,38 @@
...
+initClient();
+initClientNavigation();
+```
+
+Once this is initialized, internal `<a href="/some-path">` links will no longer trigger full-page reloads. Instead, the SDK will:
+
+1. Intercept the link click,
+2. Push the new URL to the browser's history,
+3. Fetch the new page's RSC payload from the server,
+4. And hydrate it on the client.
```
justinvdm, line 24: Nice!

sdk/src/runtime/clientNavigation.ts
```diff
@@ -0,0 +1,65 @@
...
+        return;
+      }
+
+      // Skip if target="_blank" or similar
+      if (link.target && link.target !== "_self") {
+        return;
+      }
+
+      // Skip if download attribute
+      if (link.hasAttribute("download")) {
```
justinvdm, line 42: Should we document this?

sdk/src/runtime/clientNavigation.ts
```diff
@@ -0,0 +1,65 @@
+export function initClientNavigation(
+  opts: {
+    onNavigate: () => void;
+  } = {
+    onNavigate: async function onNavigate() {
+      // @ts-expect-error
+      await globalThis.__rsc_callServer();
```
justinvdm, line 7: Incredible how simple this is!

justinvdm (approved):
:shipit: 🚀

sdk/src/runtime/clientNavigation.ts
peterp: My intuition says that it's such a little footnote that it doesn't really matter. It's the expected behavior.

sdk/src/runtime/clientNavigation.ts
peterp: Yeah, it's pretty incredible.

peterp:
At some point, I would like to get a hook on the fetch that has been called by call server because then I can actually implement some kind of preload functionality.

peterp:
I just realized that if we push down a redirect response or an error response, we should probably not try and deal with that. We should just reload the browser at this point.

PR #535, opened 2025-06-21
docs: clarify ssr render flag

jldec:
I liked @justinvdm's [loom](https://www.loom.com/share/0ba7ca7d375c403fb10959e28f7ede41?sid=131f0723-b1ec-4670-9aae-72a78d830e5a) so much that I thought including that detail in the docs might help other users understand the distinction between disabling server-side rendering of client components vs. server-side rendering of server-only components.


justinvdm:
Thank you @jldec! I should've actually updated this.

jldec:
ship, improve, rinse and repeat.

PR #547, opened 2025-06-27
fix: use npx for wrangler commands to ensure correct version

Checkm3out:
Line 327 already used npx - this change updates the rest of the wrangler commands to also use npx

The issue that prompted this change was i could use npm run release but not pnpm release

After this modification pnpm release works as expected

### DIFF
--- a/sdk/src/scripts/ensure-deploy-env.mts
+++ b/sdk/src/scripts/ensure-deploy-env.mts
@@ -122,14 +122,14 @@ export const ensureDeployEnv = async () => {
     // todo(justinvdm): this is a hack to force the account selection prompt,
     // we need to find a better way
     if (!(await pathExists(accountCachePath))) {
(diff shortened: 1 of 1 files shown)


peterp:
Interesting, thank you! I guess it was trying to use a globally scoped executable?

Checkm3out:
> Interesting, thank you! I guess it was trying to use a globally scoped executable?

Yes, wrangler does not recommend to install globally but i did by accident at some point - and the script was using that one until this change was applied

justinvdm:
Thank you for this @Checkm3out.

I'm surprised `pnpm release` doesn't work - when you run pnpm or npm scripts, I think they include the local `node_modules/.bin` in the `$PATH`, so wrangler should be accessible.

All the same, using `npx` throughout totally makes sense!

PR #548, opened 2025-06-27
fix: handle Ctrl+C gracefully during deployment prompt

Checkm3out:
Add SIGINT handler to properly exit when user presses Ctrl+C
instead of continuing with deployment

## Problem
When prompted "Do you want to proceed with deployment? (y/N):", pressing Ctrl+C doesn't properly cancel the deployment. The script continues running instead of exiting.

## Solution
Added a SIGINT event handler to the readline interface that:
- Closes the readline interface
- Prints "Deployment cancelled." message
- Exits with code 1

This ensures Ctrl+C behaves as expected and cancels the deployment process.

### DIFF
--- a/sdk/src/scripts/ensure-deploy-env.mts
+++ b/sdk/src/scripts/ensure-deploy-env.mts
@@ -35,6 +35,14 @@ const promptForDeployment = async (): Promise<boolean> => {
   });
 
   return new Promise((resolve) => {
+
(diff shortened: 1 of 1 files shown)


peterp:
Perfect, thank you!

justinvdm:
Released in https://github.com/redwoodjs/sdk/releases/tag/v0.1.4 :)

PR #557, opened 2025-07-01
docs: Add basic docs on DO databases

justinvdm:


barakcodes:
I can not express how much I have been waiting for something like this I even prototyped something [similar](https://github.com/barakcodes/orm-do)   can't wait to try it out

justinvdm:
Thank you @barakcodes, really glad to hear this! Keen to hear how you find `rwsdk/db`.

`orm-do` looks very cool, looks like you've explored this idea quite a bit already. It may be there's things we can take from it in `rwsdk/db`

barakcodes:
> Thank you @barakcodes, really glad to hear this! Keen to hear how you find `rwsdk/db`.
> 
> `orm-do` looks very cool, looks like you've explored this idea quite a bit already. It may be there's things we can take from it in `rwsdk/db`

will do! I have an a project in mind, will provide feedback as I go. I hit a few issues with the types with multiple dbs but I suspect that was fixed.

PR #574, opened 2025-07-06
fix @prisma types reference in tutorial

shanekunz:
So far, the tutorial has been great. There was one section that I think introduced the shared links file too early, but that's easy enough to ignore. This one tripped me up a bit so I felt compelled to make it easier.


ahaywood:
Thanks for the feedback and taking the time to submit the PR.

ahaywood (approved):

PR #581, opened 2025-07-08
fix ts param type on 7-contacts.mdx

shanekunz:
I'm not a TS expert, but I had to do this to compile. It seems `ctx` is a field inside `RequestInfo` object, not directly itself a `RequestInfo` object.


peterp:
You're right, thanks Shane!

PR #590, opened 2025-07-11
fixing the path annotation on the code block from `.../auth/Login.tsx` to `.../user/Login.tsx`

nikolasburk:
fixing the path annotation on the code block


peterp:
Thanks!

PR #591, opened 2025-07-11
Update 3-database-setup.mdx

nikolasburk:
not super sure but since the tutorial is using pnpm, should this be used to run `prisma generate` here as well?


peterp:
Yup, thanks!

PR #594, opened 2025-07-14
Remove NODE_ENV for Windows users.

peterp:


justinvdm:
@peterp I think we actually default to production - so this alone is likely not enough. I'll add to this PR for making it so that we default to development for dev and preview commands in a bit. Can you keep it open till then?

justinvdm:
> @peterp I think we actually default to production - so this alone is likely not enough. I'll add to this PR for making it so that we default to development for dev and preview commands in a bit. Can you keep it open till then?

As it turns out, vite do this already: https://github.com/vitejs/vite/blob/ae91bd0ad10942898c3d7aa8181249fb9682a4fe/packages/vite/src/node/config.ts#L1056

Except they have the defaults as this (which makes sense):
* `dev` => `development`
* `preview` => `production`
* `build` => `production`

PR #603, opened 2025-07-16
Add way to interrupt response handling for client side navigation.

peterp:
When client-side navigation is initialized, we use fetch to request a new RFC payload. This RSC payload can also be an invalid response. 
In which case, client-side navigation breaks. This PR includes the ability to manually handle the response and determine whether it's an error, and then to just forward the browser directly to that page so that the user gets the proper experience. 

```tsx
const { handleResponse } = initClientNavigation()
initClient({ handleResponse })
```

**edit(justinvdm, 22 Jul 2025):**
This PR also now includes the changes needed to get this same functionality to work with the realtime: interruptors + client navigation + realtime client:

```tsx
const { handleResponse } = initClientNavigation()
initRealtimeClient({ handleResponse })
```

The main changes on the realtime side:
* refactoring
  * to avoid repetition on worker and client side: `ACTION_*` messages (when a client triggers an action and we're getting the corresponding response) and `RSC_*` messages (when server sends up an update - e.g. as a result of a different client's triggering an action) now reuse logic where possible
  * packing and unpacking protocol messages now encapsulated in `packMessage` + `unpackMessage` + types for protocol messages
* This refactoring allowed to make way for
  * including a status code to use to rehydrate `Response`s over websocket on client side - without refactoring this turned out to be surprisingly difficult
  * for both `ACTION_*` and `RSC_*` messages, allow for short-circuiting processing of responses using `handleResponses` - to integrate with client navigation

Fixes #601

PR #605, opened 2025-07-17
fix: follow server redirects in client navigation

### LINKED ISSUE #601: Server redirects break client navigation
I tried turning on client side routing with:

```tsx
import { initClient, initClientNavigation } from "rwsdk/client";

initClient();
initClientNavigation();
```

It works in isolation. However, when any `route()` includes a redirect or an interruptor returns a redirect, the connection is closed and the client never gets hydrated, leaving the user with a blank page until they refresh.

It's easy to repro:

1. Set up a fresh app with `npx degit redwoodjs/sdk/starters/standard#main csr-issue`
2. Add `initClientNavigation` to `client.tsx`
3. Set up three `route`s, each with anchor tags pointing to all of them for easy observation:
  - Two simple static ones that returns a component/JSX
  - A third that returns a redirect to one of the others

```tsx
// worker.tsx

export default defineApp([
  setCommonHeaders(),
  render(Document, [
    route("/", () => {
      return (
        <div>
          home
          <a href="/">home</a>
          <a href="/one">one</a>
          <a href="/two">two</a>
        </div>
      );
    }),
    route("/one", () => {
      return (
        <div>
          one
          <a href="/">home</a>
          <a href="/one">one</a>
          <a href="/two">two</a>
        </div>
      );
    }),
    route("/two", () => {
      const destination = Math.random() > 0.5 ? "/" : "/one";
      return new Response(null, {
        status: 302,
        headers: {
          Location: destination,
        },
      });
    }),
  ]),
]);
```

You'll see the error and the RSC response stream drop in the console when you click on link `two`:
```
[Error] Unhandled Promise Rejection: Error: Connection closed. (clientNavigation.js 47)
	(anonymous function) (rwsdk_client.js:201)
```

```
[Error] Uncaught error: Error: Connection closed. (client.js 64)

Component stack:
Content@http://localhost:5173/node_modules/.vite/deps/chunk-IYV4OMKJ.js:703:56
	onUncaughtError (chunk-IYV4OMKJ.js:710)
	logUncaughtError (react-dom_client.js:6297)
	runWithFiberInDEV (react-dom_client.js:977)
	(anonymous function) (react-dom_client.js:6325)
	callCallback (react-dom_client.js:4797)
	commitCallbacks (react-dom_client.js:4809)
	runWithFiberInDEV (react-dom_client.js:977)
	commitLayoutEffectOnFiber (react-dom_client.js:9097)
	flushLayoutEffects (react-dom_client.js:11268:287)
	commitRoot (react-dom_client.js:11173)
	commitRootWhenReady (react-dom_client.js:10601)
	performWorkOnRoot (react-dom_client.js:10546)
	performWorkOnRootViaSchedulerTask (react-dom_client.js:11721)
	performWorkUntilDeadline (react-dom_client.js:36)
```

alecmarcus:
Closes #601

- Client navigation is now expected to be initialized with `initClient({ spaMode: true })`
- This flag is used to alter the behavior of fetch transport to follow redirects in the response recursively
- For backwards compatibility, adds a check to avoid initializing client navigation more than once
- Update the type of `onNavigate` to be a promise

I dug into this without seeing #603 😅. But if I understand correctly, my approach is different and as a result, client side rendering remains possible across same origin redirects.

I also figured that given cross-origin server redirects are acceptable with server navigation, it would make sense to gracefully fall back to server redirects when we encounter them. This is now done by catching the CORS error and falling back to browser navigation.

Unfortunately there's no way to see the location of a cross origin redirect in a fetch response, so the check is for a generic `TypeError: Failed to fetch`. However the fallback behavior is simply to reload the window at the originally requested url on the site's origin, so I don't see this being an issue.

This is working nicely from my perspective as a dev user. But if this doesn't make sense as a contribution I'm happy to make changes.

https://github.com/user-attachments/assets/df004862-8a76-4b99-8f78-d96a7efe873a

### DIFF
--- a/sdk/src/runtime/client.tsx
+++ b/sdk/src/runtime/client.tsx
@@ -1,6 +1,7 @@
 import { clientWebpackRequire } from "./imports/client";
 import { type CallServerCallback } from "react-server-dom-webpack/client.browser";
 import { type HydrationOptions } from "react-dom/client";
+import { initClientNavigation } from "./clientNavigation";
 
 // NOTE: `react-server-dom-webpack` uses this global to load modules,
 // so we need to define it here before importing "react-server-dom-webpack."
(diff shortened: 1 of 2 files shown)


justinvdm:
Thank you for this @alecmarcus!

We've released the fix to #601 here: https://github.com/redwoodjs/sdk/releases/tag/v0.1.24

Now to figure out where to go for this PR here. It looks like we took quite different directions, and so my take is: lets figure out if there's anything here that we haven't covered in #603, so we can figure out what to do about it.

Along those lines, there was mention of cross origin redirects. @peterp Do you think this is handled already with #603 now landed?

Anything else covered here and not in #603?

alecmarcus:
My pleasure!

As I understand it, the main and only difference is that this approach gives continuous client side rendering for all same-origin requests. You can see in the video that instead of falling back to a full reload via browser navigation, we're able to maintain the diff-and-rerender SPA behavior for any number of redirects the server might respond with, whereas I believe #603 immediately falls back to a full page reload at the next location when the server returns a redirect.

That being the case, #603 should handle cross origin request just fine/the same way this change does

peterp:
Hey @alecmarcus Thanks for opening this; it helped a lot with the implementation, but since the functionality has landed I've closed it for now. Please ping me if I missed something?

PR #609, opened 2025-07-21
fix: NODE_ENV=development deploys and previews

### LINKED ISSUE #607: Client-side error after running build + preview or release with NODE_ENV=development
I was tripped up today because my local environment had `NODE_ENV=development`
This caused client-side errors e.g. with `vite build && vite preview` and on deployments made with `pnpm run release`.  

To reproduce see https://github.com/jldec/create-minimal-jul-20 or

- create a project with `pnpm create rwsdk -t minimal`
- add a [Button](https://github.com/jldec/create-minimal-jul-20/blob/main/src/app/pages/Button.tsx) component with 'use client' to the Home page.
- open a preview build with `NODE_ENV=development pnpm vite build && pnpm vite preview` OR
- deploy using NODE_ENV=development pnpm release

The client-side error occurs because of a request for `Button.tsx` which returns a 404 with `content-type: text/html`

Screenshot 2025-07-20 at 13 22 11

justinvdm:
## Context
So far, we've (I've) been been using `import.meta.DEV` to determine both:
* Is the the dev server running
* Is the application running in dev mode (`NODE_ENV=development`)

This is wrong of course, and breaks down in cases where one wants to use `NODE_ENV=development` in deployments or previews.

Why would one want to use `NODE_ENV=development` for deploys or previews? Use cases include easier debugging more dev info on staging deployments and vite previews.

## Solution
We now no longer conflate the dev server running with whether `NODE_ENV=development` was given. We now have:
* `import.meta.env.VITE_IS_DEV_SERVER` - only true when the dev server is running (not a deployment or preview). Is independent of the value of `NODE_ENV=development` or the mode that vite is running in
* `import.meta.env.DEV` - determined by the mode that vite is running in - we still have this via vite, and it still implies the app is running in development mode, which defaults to whether was `NODE_ENV=development`

We now use `VITE_IS_DEV_SERVER` in places where what we are actually trying to determine, is whether the app is running via the dev server. For example:
* when determining whether the load modules dynamically via the dev server or not
* when determining whether to inject the vite dev HMR preamble

Fixes #607

### DIFF
--- a/docs/src/content/docs/tutorial/full-stack-app/2-create-app.mdx
+++ b/docs/src/content/docs/tutorial/full-stack-app/2-create-app.mdx
@@ -261,12 +261,11 @@ Before this will work in production, we need to update our security headers to a
 
 ```tsx title="src/app/headers.ts" collapse={1-26} {30}
 import { RouteMiddleware } from "rwsdk/router";
-import { IS_DEV } from "rwsdk/constants";
 
 export const setCommonHeaders =
 (): RouteMiddleware =>
 ({ headers, rw: { nonce } }) => {
-  if (!IS_DEV) {
+  if (!import.meta.env.VITE_IS_DEV_SERVER) {
     // Forces browsers to always use HTTPS for a specified time period (2 years)
     headers.set(
       "Strict-Transport-Security",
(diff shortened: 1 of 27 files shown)


justinvdm:
Fixed in https://github.com/redwoodjs/sdk/releases/tag/v0.1.21 :)

PR #611, opened 2025-07-21
fix: renderRealtimeClients() usage with client navigation

justinvdm:
## Problem
For each realtime channel, we currently store the relevant url for each connected client, and use this when rendering clients on demand using `renderRealtimeClients()`. However, we weren't accounting for changes to the url done as a result of client navigation.

## Solution
We already use each url we get for a client when client actions and navigations happen to determine how the client should be updated. We just need to store this latest url as the new url for usage with `renderRealtimeClient()`.


jldec:
I ❤️ how simple the fix for this was - validates the pjax design for client-side navigation.

PR #616, opened 2025-07-27
feat: add native scroll behavior to client navigation

oscabriel:
**Problem:** initClientNavigation() did not define any scroll behavior, meaning that every type of navigation caused scroll position to persist between page navigations, which feels unnatural as a default.

**Solution:** Added configurable scroll behavior to initClientNavigation() that mirrors the native default behavior for MPAs.

**Changes:**
- New options interface:
  - Added scrollToTop?: boolean (default: true) - controls whether to scroll to top after navigation
  - Added scrollBehavior?: "auto" | "smooth" | "instant" (default: "instant") - controls scroll animation
  - Updated options interface to work when setting scrollToTop and scrollBehavior but not onNavigate

- Enhanced scroll management:
  - Disables the browser's automatic scrollRestoration to prevent visual shift when navigating
  - Implements manual scroll position save/restore for popstate events
  - Saves scroll position before navigation and restores it on back/forward nav
  - Uses MutationObserver to trigger scroll behavior after DOM updates

- Improved history management:
  - Saves current scroll position in history state before navigation
  - Updates history state with new scroll position after navigation
  - Ensures consistent scroll behavior when navigating back and forward
  
- Updated documentation for client navigation

### DIFF
--- a/docs/src/content/docs/guides/frontend/client-side-nav.mdx
+++ b/docs/src/content/docs/guides/frontend/client-side-nav.mdx
@@ -9,11 +9,11 @@ Client-side navigation is a technique that allows users to move between pages wi
 
 This approach is commonly referred to as a Single Page App (SPA). In RedwoodSDK, you get SPA-like navigation with server-fetched React Server Components (RSC), so it's fast and dynamic, but still uses the server for rendering.
 
-```tsx title="src/client.tsx" ins="initClientNavigation()"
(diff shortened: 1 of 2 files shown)


peterp:
LGTM! Thanks @oscabriel

justinvdm:
Released in https://github.com/redwoodjs/sdk/releases/tag/v0.1.27

PR #620, opened 2025-07-29
shadcn docs: path update, note combination, steps rearrange

pantheredeye:
This PR updates a few things on the ShadCN page. It began as a way to make it clear that the lib folder should be moved into the app folder. After making that note change, a few other things stood out that were easy to clean up.

* The order of steps is changed. Install Tailwind is step 1 instead of 3, because shad install will fail if tailwind check fails.
* The notes regarding the lib folder are combined into a single note. 
* Made it clear that, if using Redwood conventions, the aliases should match the config.
* Corrected a path error stating where components will be added.

Original on Left; Updated on the Right. 

Screenshot_29-7-2025_16154_docs


ahaywood:
Thanks @pantheredeye! I made a few more tweaks and merged this branch in #710

Issue #569, opened 2025-07-05
initClient fails when loading 'new Response(await renderToStream(< />, { Document })'


jldec:
I think there are missing RSC headers or something like that when you return a `renderToStream()` document in the response, instead of letting the router handle it. This happens for status 200 as well (not just status 404).

#### Repro
- git clone https://github.com/jldec/agents-chat
- git checkout 56a5dcaa
- modify client.tsx not to ignore pages with element id '404'
  e.g.   `if (!document.getElementById('404_IS_NOT_THE_ID_')) {`
- pnpm install, pnpn dev
- navigate to an unknown path e.g. http://localhost:5173/foo

```
Uncaught error:  Error: An unsupported type was passed to use(): undefined
```

<img width="1135" height="782" alt="Image" src="https://github.com/user-attachments/assets/becbee13-9b3a-4c11-adcf-8546ed2c8f1d" />

<img width="1133" height="790" alt="Image" src="https://github.com/user-attachments/assets/8003dcf5-510c-40c4-a5e8-ffac6f641919" />

PR #630, opened 2025-07-31
fix: Handling 404 responses with client side navigation

justinvdm:
### Problem

The framework's routing API allows a route to return a React element, a `Response` object, or a promise that resolves to either. To handle this, a route is executed as a function inside a wrapper component. If the function returns a `Response`, it is handled via a rejection that is caught by the request handler to break out of the React rendering cycle.

This process encounters a timing issue when a route performs asynchronous work, such as fetching data or calling `renderToStream`, before returning a value. The request handler may begin sending a streamed response to the client before the route's promise has resolved. If the promise later resolves to a `Response`, the rejection may occur *after* the response stream has already started, making it impossible to switch from the streamed content to the returned `Response`. This breaks features that rely on intercepting requests, like redirects or 404s, especially during client-side navigation.

### Solution

This change resolves the timing issue by using a deferred created with `Promise.withResolvers()` to synchronize the route's resolution with the response stream.

When a route is rendered, the request handler awaits the promise from this deferred before sending the final response.

- If the route's logic resolves with a React element, the deferred is resolved, allowing the rendering stream to proceed as normal.
- If the route's logic resolves with a `Response` object, the deferred is rejected with that `Response`. This rejection is caught before the stream is sent, allowing the `Response` to be handled correctly and sent to the client.

### DIFF
--- a/sdk/src/runtime/lib/router.ts
+++ b/sdk/src/runtime/lib/router.ts
@@ -222,14 +222,16 @@ export function defineRoutes<T extends RequestInfo = RequestInfo>(
           if (isRouteComponent(h)) {
             const requestInfo = getRequestInfo();
             const WrappedComponent = wrapWithLayouts(
-              h as React.FC,
+              wrapHandlerToThrowResponses(h as RouteComponent<T>) as React.FC,
               layouts || [],
               requestInfo,
             );
-            // context(justinvdm, 31 Jul 2025): We now know we're dealing with a page route,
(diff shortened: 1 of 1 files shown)


jldec:
I think this resolves #569 as well :)

justinvdm:
Released in
* https://github.com/redwoodjs/sdk/releases/tag/v0.1.28
* https://github.com/redwoodjs/sdk/releases/tag/v0.2.0-alpha.1

PR #631, opened 2025-07-31
Make scroll restoration declarative.

peterp:
Previous iteration of scroll restoration relied on mutation observation API. I was not certain when the mutation observation event would fire, so I made this declarative rather than a side effect.

I've also added the option to disable automatic scroll behavior using `history.scrollRestoration = manual`


justinvdm:
Released in:
* https://github.com/redwoodjs/sdk/releases/tag/v0.1.30
* https://github.com/redwoodjs/sdk/releases/tag/v0.2.0-alpha.3

PR #633, opened 2025-08-01
Allow response status and headers to be modified.

peterp:
I think that we need to consider adding this as an option. I don't know if this is the right way to go about it, but I wanted to start the conversation.

The `renderToString` and `renderToStream` methods are, I think, quite difficult to get right - and this is a quick win for us so that we can unblock @jldec.

I think I found better names for the request info object. I think we should call it HTTP messages. Which is what the Firefox manual refers to as request and response. I think we should also deprecate the headers and then just require that people set it on the actual response object.

### DIFF
--- a/sdk/src/runtime/requestInfo/types.ts
+++ b/sdk/src/runtime/requestInfo/types.ts
@@ -2,11 +2,17 @@ import { RwContext } from "../lib/router";
 
 export interface DefaultAppContext {}
 
+// NOTE: (peterp, 2025-08-01):
+// I think we should rename this to something other than request-info.
+// Since it's both the request and the response.
+// HttpMessages?
 export interface RequestInfo<Params = any, AppContext = DefaultAppContext> {
   request: Request;
   params: Params;
   ctx: AppContext;
+  /** @deprecated: Use `response.headers` instead */
   headers: Headers;
   rw: RwContext;
   cf: ExecutionContext;
+  response: ResponseInit;
 }
--- a/sdk/src/runtime/worker.tsx
+++ b/sdk/src/runtime/worker.tsx
@@ -72,6 +72,11 @@ export const defineApp = <
           pageRouteResolved: undefined,
         };
 
+        const userResponseInit: ResponseInit = {
+          status: 200,
+          headers: new Headers(),
+        };
+
         const outerRequestInfo: RequestInfo<any, T["ctx"]> = {
           request,
           headers: userHeaders,
@@ -79,6 +84,7 @@ export const defineApp = <
           params: {},
           ctx: {},
           rw,
+          response: userResponseInit,
         };
 
         const createPageElement = (
@@ -135,10 +141,13 @@ export const defineApp = <
           });
 
           if (isRSCRequest) {
+            const responseHeaders = new Headers(userResponseInit.headers);
+            responseHeaders.set("content-type", "text/x-component; charset=utf-8");
+            
             return new Response(rscPayloadStream, {
-              headers: {
-                "content-type": "text/x-component; charset=utf-8",
-              },
+              status: userResponseInit.status,
+              statusText: userResponseInit.statusText,
+              headers: responseHeaders,
             });
           }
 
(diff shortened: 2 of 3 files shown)


sdk/src/runtime/requestInfo/types.ts
```diff
@@ -2,11 +2,17 @@ import { RwContext } from "../lib/router";
 
 export interface DefaultAppContext {}
 
+// NOTE: (peterp, 2025-08-01):
+// I think we should rename this to something other than request-info.
+// Since it's both the request and the response.
+// HttpMessages?
```
jldec, line 8: It took my brain some time, but I now think of this as a live global interface for state associated with the current request. Since the response is also associated with the original request, I don't see a problem with 'requestInfo'.

jldec:
I tried this out in https://github.com/fmctraining/fmc-website/pull/36 and the code is much cleaner. 

Deployed with client-side nav at https://response-status-fmc-website.fmctraining.workers.dev/

Look for "Broken Link" on home page or go to   
https://response-status-fmc-website.fmctraining.workers.dev/this/is/a/broken-link/

The only issue I noticed was a type mismatch for `Document` in worker.tsx
```
Argument of type 'FC<{ children: ReactNode; }>' is not assignable to parameter of type 'FC<DocumentProps<RequestInfo<any, DefaultAppContext>>>'.
  Type 'React.ReactNode | Promise<React.ReactNode>' is not assignable to type 'import("/Users/jldec/cloudflare/redwood/sdk/node_modules/.pnpm/@types+react@19.1.2/node_modules/@types/react/index").ReactNode | Promise<import("/Users/jldec/cloudflare/redwood/sdk/node_modules/.pnpm/@types+react@19.1.2/node_modules/@types/react/index").ReactNode>'.
    Type 'ReactElement<unknown, string | JSXElementConstructor<any>>' is not assignable to type 'ReactNode | Promise<ReactNode>'.
      Property 'children' is missing in type 'ReactElement<unknown, string | JSXElementConstructor<any>>' but required in type 'ReactPortal'.
```

sdk/src/runtime/worker.tsx
```diff
@@ -72,13 +72,19 @@ export const defineApp = <
           pageRouteResolved: undefined,
         };
 
+        const userResponseInit: ResponseInit = {
+          status: 200,
+          headers: new Headers(),
+        };
```
justinvdm, line 78: It didn't occur to me we could use `ResponseInit` type and maintain object instead of trying to maintain a `Response` object 💡 

Means we can keep things mutable, while still following the standards. Nice!

justinvdm:
> I tried this out in [fmctraining/fmc-website#36](https://github.com/fmctraining/fmc-website/pull/36) and the code is much cleaner.
> 
> Deployed with client-side nav at https://response-status-fmc-website.fmctraining.workers.dev/
> 
> Look for "Broken Link" on home page or go to https://response-status-fmc-website.fmctraining.workers.dev/this/is/a/broken-link/
> 
> The only issue I noticed was a type mismatch for `Document` in worker.tsx
> 
> ```
> Argument of type 'FC<{ children: ReactNode; }>' is not assignable to parameter of type 'FC<DocumentProps<RequestInfo<any, DefaultAppContext>>>'.
>   Type 'React.ReactNode | Promise<React.ReactNode>' is not assignable to type 'import("/Users/jldec/cloudflare/redwood/sdk/node_modules/.pnpm/@types+react@19.1.2/node_modules/@types/react/index").ReactNode | Promise<import("/Users/jldec/cloudflare/redwood/sdk/node_modules/.pnpm/@types+react@19.1.2/node_modules/@types/react/index").ReactNode>'.
>     Type 'ReactElement<unknown, string | JSXElementConstructor<any>>' is not assignable to type 'ReactNode | Promise<ReactNode>'.
>       Property 'children' is missing in type 'ReactElement<unknown, string | JSXElementConstructor<any>>' but required in type 'ReactPortal'.
> ```

Will take a look! 👍 

If I'm following, that one isn't not directly related to this PR, but more now a consequence of the fact that we're now able to return the element from the interruptor directly since we can now set response status instead of needing to through `renderToStream` hoop. With that in mind, I'll merge and release this PR so long, but look at this type issue next.

justinvdm:
Released in:
* https://github.com/redwoodjs/sdk/releases/tag/v0.2.0-alpha.2
* https://github.com/redwoodjs/sdk/releases/tag/v0.1.29

PR #648, opened 2025-08-13
fix: HMR for SSR'd css modules

justinvdm:
## Context
A unique challenge arises from how stylesheet imports are processed in the development server's SSR environment. When a client component imports a CSS file, that import travels through the [SSR Bridge](./ssrBridge.md), which must correctly handle it. If not managed properly, plugins like Tailwind CSS might try to parse transformed JavaScript (from a CSS module) as if it were raw CSS, causing build failures.

This problem is orthogonal to injecting stylesheet links into the `<head>`. It is about ensuring that `import styles from './Button.module.css'` works correctly during server-side rendering, providing the component with the correct class name hashes for the initial HTML.

To solve this, we performs a specific transformation when processing for the ssr bridge:

1.  **Identifier Transformation:** When resolving a module ID for a stylesheet (e.g., `/src/components/Button.module.css`), the plugin appends a `.js` extension to its virtualized path (e.g., `virtual:rwsdk:ssr:/src/components/Button.module.css.js`). This signals to all subsequent Vite plugins that the module should be treated as JavaScript.

2.  **Specialized Loading:** In the `load` hook, the plugin distinguishes between two types of CSS imports:
    -   For **standard `.css` files**, it returns an empty JavaScript module (`export default {};`). This is because global stylesheets are handled by Vite on the client and do not need to be processed as part of the server-rendered component tree.
    -   For **CSS Modules (`.module.css`)**, it first strips the `.js` suffix from the module ID before calling `devServer.environments.ssr.fetchModule()`. This ensures the original CSS module is fetched and transformed by Vite into its JavaScript object representation. This object, which maps class names to their unique generated hashes, is then returned to the `worker` environment so that server-rendered components receive the correct class names.

## Problem + Solution
We were adding this .js extension for the SSR bridge, but during HMR invalidation, we were not accounting for this added extension. We just needed to add the extension there too so that what we try invalidate for HMR matches what is in the module graph.

### DIFF
--- a/sdk/src/vite/miniflareHMRPlugin.mts
+++ b/sdk/src/vite/miniflareHMRPlugin.mts
@@ -123,6 +123,7 @@ export const miniflareHMRPlugin = (givenOptions: {
           type: "full-reload",
           path: "*",
         });
+        log("hmr: Full reload after error");
         return [];
       }
       const {
@@ -144,6 +145,7 @@ export const miniflareHMRPlugin = (givenOptions: {
       }
 
       if (!isJsFile(ctx.file) && !ctx.file.endsWith(".css")) {
+        log(`hmr: not a js file, skipping`);
         return;
       }
 
@@ -156,6 +158,7 @@ export const miniflareHMRPlugin = (givenOptions: {
         });
 
         if (!isUseClientUpdate) {
+          log("hmr: not a use client update, short circuiting");
           return [];
         }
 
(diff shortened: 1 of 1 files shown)


BrendanBerkley:
Just want to say: it feels good to come across a few problems while developing with this and then find a couple of releases that fix the problems. Thanks for your continued efforts on getting CSS modules working!

PR #660, opened 2025-08-25
add an iframe to starter page so it looks more refined

harryhcs:
Update Document component and security headers
- Added CSS module and module preload link to the Document component for iframe styling
- Updated Content Security Policy in headers to allow frame ancestors from 'self' and local dev
- Replaced static content in Home component with an iframe to load external start page.


justinvdm (approved):
This way we get onboarding wins and telemetry without bloating the starter, clever!

justinvdm:
@harryhcs i'll release this tomorrow morning when we're both around, ~9am, if that is alright?

PR #715, opened 2025-09-10
fix: Unify request handling for pages and RSC actions

justinvdm:
### Problem

Previously, RSC action requests bypassed the middleware pipeline and were handled directly. This created an inconsistency where logic defined in middleware (e.g. session validation, authentication checks, setting security headers) was not applied to action requests. This could lead to security vulnerabilities and required developers to duplicate context-aware logic inside action handlers.

### Solution

The router's `handle` method has been restructured into a deterministic, three-stage pipeline that processes all incoming requests:

1.  **Global Middleware Execution**: All global middleware is executed in order.
2.  **RSC Action Handling**: If the request is an RSC action, the action handler is now invoked *after* all middleware has completed.
3.  **Page Route Matching**: The router proceeds to match the request to a page route for rendering.

### DIFF
--- a/.worklogs/justin/2025-09-10-rsc-actions-in-middleware.md
+++ b/.worklogs/justin/2025-09-10-rsc-actions-in-middleware.md
@@ -1,0 +1,35 @@
+# RSC Actions in Middleware
+
+## Problem
+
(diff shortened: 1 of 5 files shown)


peterp:
This is working well for me!

PR #718, opened 2025-09-11
fix: Prevent module state loss from dev server re-optimization

justinvdm:
### Problem

After adding our [dep optimization scan phase](https://github.com/redwoodjs/sdk/blob/4d3ebb28f0910cae022eaa4462d62562422583c3/docs/architecture/directiveScanningAndResolution.md), we no longer provided all client and server files as entry points to Vite's dependency optimizer (`optimizeDeps`). Previously, this allowed Vite to discover all dependencies ahead of time.

Without this list of entry points, dependencies are now discovered at runtime as the relevant client and server files are requested. When Vite encounters a new dependency this way, it triggers a re-optimization pass, which reloads the server worker. This reload creates new instances of all modules, resetting any module-level state. As a result, features that rely on this state, such as the `requestInfo` object which uses `AsyncLocalStorage`, would break and cause operations to fail.

### Solution

This change extends our "barrel file" strategy to inform Vite's dependency optimizer of all application entry points without degrading dev server startup performance ([context here](https://github.com/redwoodjs/sdk/blob/c33eb0f2adae3dae64b92cfefcac7c7897bd9700/docs/architecture/devServerDependencyOptimization.md)).

The core of the problem is a timing issue. Our directive scan, which finds all client and server files, runs asynchronously. By the time it completes, Vite's `optimizeDeps` process has already started and is no longer reading from its initial configuration, making it too late to add our discovered files to the `entries` array. The alternative would be to block the `optimizeDeps` step entirely, which would also block the dev server from starting.

This solution uses a custom `esbuild` plugin to hook into the `optimizeDeps` process as it runs:
1.  We provide the paths to initially empty barrel files in the `optimizeDeps.entries` configuration from the start.
2.  The `esbuild` plugin's `onResolve` hook intercepts requests for these specific barrel files.
3.  When a barrel file is requested by the optimizer, the hook awaits our asynchronous directive scan. Once the scan is complete, the barrel files are populated with the full list of application entry points.
4.  The hook then allows the optimizer to proceed with the now-populated barrel files.

This approach allows the dev server to start up quickly while still correctly informing the optimizer of all necessary entry points for a complete dependency scan. This prevents the runtime discovery of new dependencies and the disruptive re-optimizations that were causing module state loss, without sacrificing startup performance.


peterp:
Awesome!

PR #786, opened 2025-09-25
fix(streaming): Ensure Early Hydration with Suspense-Aware Stream Interleaving

justinvdm:
### Context

Previously, the framework rendered the `<Document>` shell and the application content in two separate server passes. It then stitched the two resulting HTML streams together. This was done to solve a `useId` hydration mismatch by preventing the document's render from interfering with the app's render.

### Problem: Blocked Hydration with Suspense

While this approach fixed the `useId` issue, it caused a problem with Suspense. The stream stitching would wait for the entire application stream to complete before sending the rest of the document.

If the application used `<Suspense>`, React would pause the app stream to wait for data. This pause meant the rest of the document stream, including the `<script>` tag for client-side hydration, was also delayed. The UI shell would appear in the browser, but it would not be interactive until all data fetching was finished.

### Solution: Interleaving Streams via RSC Payload Markers

This change updates the stream stitching to be aware of Suspense boundaries.

The solution works by injecting a marker component (`<div id="rwsdk-app-end" />`) directly into the RSC payload on the server. Because this marker is part of the payload, it is present in both the server-side HTML render and the client-side component tree, which guarantees structural consistency and prevents hydration errors.

A utility then uses this marker in the HTML stream to interleave the document and application streams. It sends the document head, then the initial app content (up to the marker), then the *rest of the document body* (including the client script), and only then streams the suspended content from the app before finally closing the document.

This ensures the client script is sent to the browser as soon as the initial UI is ready, making the page interactive right away, without re-introducing `useId` bugs or relying on brittle implementation details of React's streaming format.

For a detailed explanation, see the updated [Hybrid Rendering documentation](https://github.com/redwoodjs/sdk/blob/6cf5885906cba0ecb97b4100e1a979e9f7c0d9ad/docs/architecture/hybridRscSsrRendering.md).


justinvdm:
re https://github.com/redwoodjs/sdk/actions/runs/18019255208/job/51272248217

The one e2e test failure is a transient/flaky test issue and it passes locally. Merging

PR #788, opened 2025-09-26
fix: Include rwsdk worker deps in ssr optimizeDeps.include

justinvdm:
Same as #785, but this time for adding `rwsdk/<worker_deps>` to SSR environment's `optimizeDeps.include`


justinvdm:
Failing E2E tests pass locally and look like flaky tests, merging and releasing,

PR #826, opened 2025-10-15
Implement HTTP-method-based routing

zshannon:
hey crew 👋 i really appreciate how responsive you all have been to support and feature requests so i wanted to contribute some code back as a gesture of my gratitude. i hope this is easy to merge :)

### DIFF
--- a/docs/src/content/docs/core/routing.mdx
+++ b/docs/src/content/docs/core/routing.mdx
@@ -23,6 +23,10 @@ export default defineApp([
   route("/ping", function handler({ request, ctx }) {
     return new Response("Pong!")
   }),
+  route("/api/users", {
+    get: () => new Response(JSON.stringify(users)),
+    post: () => new Response("Created", { status: 201 }),
+  }),
 ]);
 ```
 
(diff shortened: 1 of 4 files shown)


justinvdm:
Hey @zshannon, this looks awesome! Thank you for covering everything (implementation, tests, docs) and considering the nuanced things like custom methods and disabling methods.

This was a long-standing todo for us, appreciate it!

I want to get Peter's thoughts on the API to make sure it's inline with what we were imagining for this, though it looks great.

After that ^ and once CI green, will merge and release!

docs/src/content/docs/core/routing.mdx
```diff
@@ -111,6 +115,66 @@ Return values:
...
+
+You can handle different HTTP methods (GET, POST, PUT, DELETE, etc.) on the same path by passing an object with method keys:
+
+```tsx
+route("/api/users", {
+  get: () => new Response(JSON.stringify(users)),
+  post: ({ request }) => new Response("User created", { status: 201 }),
+  delete: () => new Response("User deleted", { status: 204 }),
+});
+```
```
justinvdm: @peterp how does this API look to you? It is also what I had in mind for when we supported methods, but also want to see what you think / how it maps to what you might've imagined for this.

justinvdm:
Closing then re-opening to get new CI workflow changes to run, sorry for the noise!

justinvdm:
Released in https://github.com/redwoodjs/sdk/releases/tag/v1.0.0-beta.15

marcomuser:
awesome job!! @zshannon would it be possible to attach the http method as well to the `RouteDefinition`? This would allow further processing of the routes, e.g. for OpenAPI spec generation (this requires more work but the http method in the RouteDefinition would be a first step).

zshannon:
> awesome job!! @zshannon would it be possible to attach the http method as well to the `RouteDefinition`? This would allow further processing of the routes, e.g. for OpenAPI spec generation (this requires more work but the http method in the RouteDefinition would be a first step).

there's an internal method:
```typescript
function isMethodHandlers<T extends RequestInfo = RequestInfo>(
  handler: RouteHandler<T> | MethodHandlers<T>,
): handler is MethodHandlers<T> {
  return (
    typeof handler === "object" && handler !== null && !Array.isArray(handler)
  );
}
```

that you can (duplicate yourself for eg OpenAPI spec generation) call on `RouteDefinition['handler']` and `if true` then `Object.keys` gives you http methods (+ a bit of fiddling if supporting non-standard http verbs via `custom` key), and `else` http methods are not statically determined

Issue #837, opened 2025-10-22
Load assets respecting Vite's `base` value


rbasto1:
Our workers are growing quite large and we have a plan to split some of the routes into separate workers.
I'm trying to have a rwsdk worker on a sub-path, ie. `www.domain.com/path`
Wrangler accepts that easily:
```
      "routes": [
        {
          "pattern": "www.domain.com/path/*",
          "zone_name": "domain.com"
        }
```
Vite as well:
```
export default defineConfig({
  base: "/path",
```
And I can have a `prefix('/path', [` added to my worker.

But when building, the path to the js assets that replace `<script>import("/src/client.tsx")</script>` are prefixed by `/` (eg. `import("/assets/client-BKZATaFV.js")`, ignoring these other configurations.

It would be great if the output html could check Vite's "base" and append it to generated paths.

peterp:
Thanks for reporting this; we will fix this very shortly.

gching:
Actually, I ran into this issue myself, and made some changes internally to `linkerPlugin` to make this work, will shoot up a PR if this is the correct implementation

gching:
@rbasto1 merged in and released at https://github.com/redwoodjs/sdk/releases/tag/v1.0.0-beta.18, try it out if it solves it for you

rbasto1:
Since I was still on 0.3.12, I might be missing some changes unrelated to this.
I've followed the steps at https://docs.rwsdk.com/migrating/, but now even when I try to deploy to a custom domain I can't make it work. I get this in the Cloudflare logs:
```
{
  "source": {
    "level": "error",
    "message": "rwsdk: Received an unhandled error:\n\n%s TypeError: Cannot read properties of undefined (reading 'H')"
  },
[...]
}
```
and
```
{
  "source": {
    "message": "Cannot read properties of undefined (reading 'H')",
    "exception": {
[...]

PR #839, opened 2025-10-23
feat: Respect Vite `base` for linking asset placeholders

gching:
Ref #837 

This PR adds in support for the Vite `base` for the linker for asset placeholders. This is needed since in the current second workers linker build, we are statically replacing asset placeholders to final paths that don't respect this `base` setting

> Not a main contributor, so this might be the wrong spot. Let me know if this logic should be elsewhere

## Changes
* Store reference to config
* Pass in `base` to `linkWorkerBundle` to prepend if it exists
* Simple tests

## Before change
With Vite `base` = `https://example.com/`
```js
// Before linking
const stylesheet = "rwsdk_asset:/src/styles.css";
const logo = "rwsdk_asset:/src/logo.svg";

// After linking
// Note not including the base
const stylesheet = "/assets/styles.123.css"
const logo = "/assets/logo.abc.svg"
```

##  After change
With Vite `base` = `https://example.com`
```js
// Before linking
const stylesheet = "rwsdk_asset:/src/styles.css";
const logo = "rwsdk_asset:/src/logo.svg";

// After linking
// Note not including the base
const stylesheet = "https://example.com/assets/styles.123.css"
const logo = "https://example.com/assets/logo.abc.svg"
```

### DIFF
--- a/sdk/src/vite/linkerPlugin.mts
+++ b/sdk/src/vite/linkerPlugin.mts
@@ -1,7 +1,7 @@
 import debug from "debug";
 import fsp from "node:fs/promises";
 import path from "node:path";
-import type { Plugin } from "vite";
+import type { Plugin, ResolvedConfig } from "vite";
 import { CLIENT_MANIFEST_RELATIVE_PATH } from "../lib/constants.mjs";
 import { normalizeModulePath } from "../lib/normalizeModulePath.mjs";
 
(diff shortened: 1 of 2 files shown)


justinvdm:
Thank you @gching, this looks solid! Merging and releasing

justinvdm:
Released in https://github.com/redwoodjs/sdk/releases/tag/v1.0.0-beta.18

Issue #840, opened 2025-10-23
[dx] Decouple `renderToStream` from `requestInfo`


gching:
## Problem
* `renderToStream` and it's underlying implementation is heavily tied to `requestInfo` and the `AsyncLocalStorage`
* This means `renderToStream` is tied to being only used in `defineApp` or more specifically `runWithRequestInfo`
* `renderToStream` only uses specific specific values from `requestInfo`, ie. `nonce`

## Current DX
* `renderToStream` can only be used in a route in `defineApp`
* `renderToStream` can only be used if ran under `runWitRequestInfo`
* Ex:
```js
  const html = await runWithRequestInfo(
    {...},
    async () => {
      return renderToString(<ResourceComp />, {
        Document,
      });
    }
  );
```

## Proposed DX Solution
* Decouple `renderToStream` and underlying implementation to not depend on the `AsyncLocalStorage` and have the bare minimum passed in as argument
* Given this, `rendeToStream` + `renderToString` does not depend on `runWithRequestInfo`
* Ex:
```js
const html = renderToString(<Component />, { Document, requestInfo: {...} })
// Or
const html = renderToString(<Component />, { Document, nonce: "..." })
```
* Benefits is that now we can use this anywhere

justinvdm:
Thank you @gching, makes sense 👍 

What I'm worried about decoupling is backwards compatibility for people that are already using `renderTo*` that are assuming they can use `requestInfo`. That, and people might expect they can use `requestInfo` with it.

What if we changed the `renderToStream` implementation to apply defaults for `requestInfo`? That way, we get best of both worlds.
* If you don't use `defineApp()`, you're none the wiser
* If you do, you get what was already applied to `requestInfo` via routing.

To do that, we could:
1. add a `runWithRequestInfoDefaults` (similar to [`runWithRequestInfoOverrides`](https://github.com/redwoodjs/sdk/blob/43aa564280cebdb2f87f6ed1ba990670f95aec34/sdk/src/runtime/requestInfo/worker.ts#L58-L70), but defaulting props rather than overriding)
2. use `runWithRequestInfoDefaults` in `renderToStream` implementation, providing it similar defaults to what we provide [here for `defineApp()`](https://github.com/redwoodjs/sdk/blob/d49a7868c99fa5bf27c439d760ad56346026b5ab/sdk/src/runtime/worker.tsx#L110-L118)

What do you think?

gching:
@justinvdm Makes sense!

[...]

PR #842, opened 2025-10-25
feat: Decouple rendering from async context 

gching:
Ref #840

The current rendering functions (`renderToStream`, `renderToString`) are tightly coupled to the **`requestInfo`** object provided by async local storage. This makes it difficult to use our rendering engine in environments outside `defineApp`.

This PR decouples the rendering logic from this implicit context.

## Changes

*   **`constructWithDefaultRequestInfo`**: Introduced and exposes a new utility function in entry from `sdk/src/runtime/requestInfo/utils.tsx`. This function creates a default `RequestInfo` object and merging with user-provided overrides.
*   **Refactored Rendering Functions**: `renderToStream` and `renderToString` were updated to accept an optional `requestInfo` parameter

### Before Change

* `renderToStream` implicitly depended on the `requestInfo` from the async local storage context
* To call these render methods, you had to import the internal `runWithRequestInfo` function and manually construct a complete `RequestInfo` object.

```typescript
  // Would error, since no requestInfo context.
  const html = await renderToString(<Welcome />);


  const requestInfo = {
    // Need to know internals and understand defaults
    // ...
  }

  await runWithRequestInfo(requestInfo, async () => {
    return renderToString(<Welcome />);
  });
```

## After Change

* Rendering functions now accept an optional `requestInfo` object
* Rendering functions automatically build up a default `requestInfo` and overrides it given the context `requestInfo` (backwards compatible) if it exists and user provided `requestInfo` overrides
* Exposes `constructWithDefaultRequestInfo` to be used to construct a `requestInfo` if needed (ie. we can use this in `defineApp` potentially)

```ts
  // Calling it without context won't error, and has defaults
  const html = await renderToString(<Welcome />);

  // Support simple overrides, ie. nonce
  const html = await renderToString(<Welcome />, {requestInfo: {nonce: 'hi-gavin-this-is-random'});

  // Can construct outside if needed in case we need a full requestInfo
  const requestInfo = {
    //...
  }
  // or
  const requestInfo = constructWithDefaultRequestInfo({...})

  // Supports passing requestInfo in it's entirety or utilize context in the past
  const html = await renderToString(<Welcome />, {requestInfo});

  await runWithRequestInfo(requestInfo, async () => {
    return renderToString(<Welcome />);
  });
  
```

### DIFF
--- a/sdk/src/runtime/entries/worker.ts
+++ b/sdk/src/runtime/entries/worker.ts
@@ -6,6 +6,7 @@ export * from "../register/worker";
 export * from "../render/renderToStream";
 export * from "../render/renderToString";
 export * from "../requestInfo/types";
+export * from "../requestInfo/utils";
 export * from "../requestInfo/worker";
 export * from "../script";
 export * from "../worker";
--- a/sdk/src/runtime/requestInfo/utils.tsx
+++ b/sdk/src/runtime/requestInfo/utils.tsx
@@ -1,0 +1,54 @@
+import { FC } from "react";
+import { type DocumentProps } from "../lib/types";
+import { generateNonce } from "../lib/utils";
+import { type PartialRequestInfo, type RequestInfo } from "./types";
+
+export const DefaultRequestInfoDocument: FC<DocumentProps> = ({ children }) => (
+  <>{children}</>
+);
+
+/**
+ * Constructs a generic requestInfo that can be used as defaults.
+ * Allows for passing in overrides to initialize with defaults.
+ */
+export const constructWithDefaultRequestInfo = (
+  overrides: PartialRequestInfo = {},
+): RequestInfo => {
+  const { rw: rwOverrides, ...otherRequestInfoOverrides } = overrides;
+
+  const defaultRequestInfo: RequestInfo = {
+    request: new Request("http://localhost/"),
+    params: {},
+    ctx: {},
+    cf: {
+      waitUntil: () => {},
+      passThroughOnException: () => {},
+      props: {},
+    },
+    response: {
+      status: 200,
+      headers: new Headers(),
+    },
+    isAction: false,
+    rw: {
+      Document: DefaultRequestInfoDocument,
+      nonce: generateNonce(),
+      rscPayload: true,
+      ssr: true,
+      databases: new Map(),
+      scriptsToBeLoaded: new Set(),
+      entryScripts: new Set(),
+      inlineScripts: new Set(),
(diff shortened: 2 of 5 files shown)


sdk/src/runtime/entries/worker.ts
```diff
@@ -6,6 +6,7 @@ export * from "../register/worker";
 export * from "../render/renderToStream";
 export * from "../render/renderToString";
 export * from "../requestInfo/types";
+export * from "../requestInfo/utils";
```
gching: Debating if we should expose this to userland, indifferent tbh

sdk/src/runtime/requestInfo/utils.tsx
```diff
@@ -0,0 +1,54 @@
+import { FC } from "react";
```
gching: Decided to make a new file for this since it contains jsx

sdk/src/runtime/requestInfo/utils.tsx
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Constructs a generic requestInfo that can be used as defaults.
+ * Allows for passing in overrides to initialize with defaults.
+ */
+export const constructWithDefaultRequestInfo = (
+  overrides: PartialRequestInfo = {},
+): RequestInfo => {
+  const { rw: rwOverrides, ...otherRequestInfoOverrides } = overrides;
+
+  const defaultRequestInfo: RequestInfo = {
```
gching: Copied over from `defineApp`, not sure if these defaults are good though

sdk/src/runtime/requestInfo/utils.tsx
```diff
@@ -0,0 +1,54 @@
...
+    request: new Request("http://localhost/"),
+    params: {},
+    ctx: {},
+    cf: {
+      waitUntil: () => {},
+      passThroughOnException: () => {},
+      props: {},
+    },
```
justinvdm: Ah, I guess we need this stubbed this way since the caller wouldn't be providing this, makes sense 👍

justinvdm:
Closing and re-opening to get CI builds to run, sorry for the noise

justinvdm:
This looks great, thank you for taking this on @gching! And I appreciate the detailed context/writeup for the PR description

Just getting the CI builds running now, will merge and release once green

justinvdm:
Released in https://github.com/redwoodjs/sdk/releases/tag/v1.0.0-beta.19

Issue #847, opened 2025-10-28
`renderToStream` outside of `defineApp` throws `__webpack_require__` not defined


gching:
## Problem
* Currently, using `renderToString` + `renderToStream` outside of `defineApp` will make `__webpack_require__` to be `undefined`
* This happens when an underlying component is a `client` component

> During dev, if there is any `defineApp`,  `renderToStream` will not throw, but in production, it will

## Reproduction
```tsx
// ServerComp.tsx
function ServerComp() {
  return <ClientComp/>
}

// ClientComp.tsx
'use client';
function ClientComp(){
  return <div/>
}


// Outside of defineApp()
const html = renderToString(<ServerComp/>, {...})
// Throws, '__webpack_require__' not defined
```

## Potential solutions
1. In `renderToStream` potentially setup the `globalThis.__webpack_require__` before proceeding
2. Expose method to userland to set `globalThis.__webpack_require__`

justinvdm:
Thanks @gching, issue makes sense. Is this something you'd be able to take on?

If yes - (1) makes sense to me (maybe we first check if it not already there though before assining). We basically need a line like this over there: https://github.com/redwoodjs/sdk/blob/d49a7868c99fa5bf27c439d760ad56346026b5ab/sdk/src/runtime/worker.tsx#L36

gching:
Yeah I think the fix should be simple, I can take it!

justinvdm:
Thank you @gching! Really appreciating your help 🙇

PR #849, opened 2025-10-29
feat: Set `__webpack_require__` + passing `requestInfo` context in `renderToStream`

gching:
ref #847 

* `renderToStream` being ran outside of `defineApp` would throw `__webpack_require__ is not defined`

## Changes
* Sets `globalThis.__webpack_require__` to `ssrWebpackRequire`
* Wraps implementation with `runWithRequireInfo` to set the requestInfo for calls

## Before Change
```tsx
// ServerComp.tsx
function ServerComp() {
  return <ClientComp/>
}

// ClientComp.tsx
'use client';
function ClientComp(){
  return <div/>
}


// Outside of defineApp()
const html = renderToString(<ServerComp/>, {...})
// Throws, '__webpack_require__' not defined
```
* Given that `__webpack_require__` isn't set, `renderToStream` will throw
## After Change

* `renderToStream` won't throw when used outside of `defineApp`

### DIFF
--- a/sdk/src/runtime/render/renderToStream.tsx
+++ b/sdk/src/runtime/render/renderToStream.tsx
@@ -55,7 +55,7 @@ export const renderToStream = async (
 
   // context(gching, 2025-10-29): We wrap the following with context to the requestInfo
   // due to `ssrWebpackRequire` needing to reference the `requestInfo` in context.
-  // Therefore, we need to wrap + also pass in the requestInfo their independent
+  // Therefore, we need to wrap + also pass in the requestInfo in their independent
   // function calls
   return runWithRequestInfo(requestInfo, async () => {
     let rscStream = renderToRscStream({


justinvdm:
Looks great, thank you @gching! Just waiting for CI, if/when green I'll merge and release.

Not a blocker for this PR at all, but I'm also wondering whether it'd be worth us having an e2e test where we're including a `"use client"` component when using `renderToStream` (we have [these e2e tests](https://github.com/redwoodjs/sdk/blob/3bf3c6fb2e676655767277b0df08868e42f8c774/playground/render-apis/__tests__/e2e.test.mts) for the `render*` apis, but that suite doesn't yet include one for `"use client"` component case). Let me know if you'd like to take that on, but 100% no pressure, I can do that too - the E2E tests are a bit of a rabbit hole. Assuming you are interested though, you can check out [`contributing.md`](https://github.com/redwoodjs/sdk/blob/114f1c12cfc0a2caa101038d9705d99004a50561/CONTRIBUTING.md#L107) for how to run and write them.

justinvdm:
Released in https://github.com/redwoodjs/sdk/releases/tag/v1.0.0-beta.22 :)

PR #860, opened 2025-11-04
Add doc strings

peterp:


sdk/src/runtime/lib/router.ts
```diff
@@ -390,6 +390,43 @@ export function defineRoutes<T extends RequestInfo = RequestInfo>(
   };
 }
 
+/**
+ * Defines a route handler for a path pattern.
+ *
+ * Supports three types of path patterns:
+ * - Static: /about, /contact
+ * - Parameters: /users/:id, /posts/:postId/edit
+ * - Wildcards: /files/*, /api/star/download
```
valscion: Should the `/star/` be `/*/` here?

```suggestion
 * - Wildcards: /files/*, /api/*/download
```

PR #862, opened 2025-11-05
fix DefaultValueExpression

xlc:
I am trying to do this and unable to. This should fix it.

```ts
.addColumn('created_at', 'integer', (col) => col.notNull().defaultTo(sql`unixepoch('now')`))
```


xlc:
Should have some tests but I am not familiar enough with the repo to know where to add it. Let me know and I can add some.

justinvdm:
Thanks @xlc! The type type tests for this live over here, if you'd like to add these: https://github.com/redwoodjs/sdk/blob/a5b5ed20db8390c76740ee85dc3e21ed52db70bb/sdk/src/runtime/lib/db/typeInference/typetests

PR #866, opened 2025-11-06
Update wildcard route pattern in documentation

valscion:
The example for wildcards should've probably used `*` character here.

Noticed this one when looking at the diff from #860


justinvdm:
Thanks @valscion!

PR #871, opened 2025-11-08
useSyncState hook.

peterp:
# Realtime Shared State Hook

This document describes the realtime synced state feature that keeps shared values aligned across browser sessions through the worker runtime.

```tsx title="src/components/Counter.tsx"
"use client"

import { useSyncState } from 'rwsdk/use-sync-state'

const Counter = () => {
  const [count, setCount] = useSyncState(0, "counter");
  return <button onClick={() => setCount((n) => n + 1)}>{count}</button>;
};
```

`setCount` applies the update locally and forwards it to the coordinator, which broadcasts the new value to every subscriber using the realtime transport.

## Challenge

Applications need a hook with the same ergonomics as `useState` that keeps values synchronised between clients while running inside the worker runtime. The solution must avoid polling, operate within the existing worker deployment model, and support multiple independent keys without leaking subscriptions.

## Overview

The design combines a Cloudflare Durable Object that stores key-value pairs, a CapnWeb RPC surface for subscription updates, and a client hook factory that wraps this RPC surface behind a React-compatible state API.

## Worker Responsibilities

- `SyncStateCoordinator` is a Durable Object that stores state values in memory, tracks subscribers per key, and broadcasts updates to connected clients via CapnWeb stubs.
- The Durable Object exposes a CapnWeb RPC target with `getState`, `setState`, `subscribe`, and `unsubscribe` methods. This keeps the RPC surface aligned with the client hook expectations.
- Worker routes proxy `/__synced-state` requests to the Durable Object using `newWorkersRpcResponse`.

## Client Hook Responsibilities

- The hook invokes `getState` for the requested key, subscribes to updates, and synchronises local state when the Durable Object pushes new values. Updates are applied optimistically in the client and forwarded through `setState`, which fans out the change to every subscriber.

## Data Flow

1. A component calls `useSyncState(initialValue, key)`.
2. The hook ensures the shared WebSocket RPC session exists and fetches the current value from the Durable Object.
3. The hook subscribes to further updates; the Durable Object records the subscriber and will invoke it on updates.
4. When any client triggers the setter, the hook sends `setState` to the Durable Object.
5. The Durable Object updates its map, iterates subscribers for the key, and invokes their RPC stubs, which push the new value to each client hook instance.
6. Each hook instance updates its local React state, keeping the UI in sync.

## Failure Handling

- Subscriber invocation failures remove the subscriber stub from the Durable Object, preventing repeated errors.
- If the Durable Object binding is missing, worker routes return a 500 response instructing the developer to restart the dev server.

## Testing

- Unit tests cover the Durable Object's subscription handling and the client hook's optimistic updates, subscription dispatch, and cleanup logic using injected dependencies.


justinvdm:
@peterp looks awesome! Should I take a look? I see its draft right now.

peterp:
I would love to merge this and silently release it! Wanna play with it a bit before we make some noise.

justinvdm:
👍 @peterp go for it! Here's how to release a new beta version: https://github.com/redwoodjs/sdk/blob/main/CONTRIBUTING.md#releasing-a-beta-version

Don't forget to merge in main first and run `pnpm install` to solve lockfile conflict. Don't worry abut re-running e2e tests after solving the conflict, I think it'd be overkill - e2e already green, just merge once conflict resolved.

valscion:
This looks pretty nice! I'd hope there's some mention of how per-user scoped synced state should be done because it wasn't obvious to me even after reading this code.

I do have a few things in mind for our upcoming app that would benefit from this hook when the time comes... 😛

peterp:
> This looks pretty nice! I'd hope there's some mention of how per-user scoped synced state should be done because it wasn't obvious to me even after reading this code.

Good point; I'll work on something

peterp:
This is what I've landed on:

The `SyncStateCoordinator.registerKeyHandler` method accepts an async function that receives a client-provided key and returns a transformed key. The handler runs in the worker request context where `requestInfo` and session data are available.

When a handler is registered, the worker routes create an RPC proxy that sits between the client and the Durable Object. For each RPC method call (`getState`, `setState`, `subscribe`, `unsubscribe`), the proxy invokes the handler to transform the key, then forwards the call to the Durable Object with the transformed key. The Durable Object operates only on transformed keys and never sees the original client-provided keys.

This approach keeps scoping logic centralized in the worker while maintaining security. The client provides simple, unscoped keys like `"counter"`, and the worker transforms them to scoped keys like `"user:123:counter"` based on the authenticated session. If the handler throws an error, the error propagates to the client.

When no handler is registered, keys pass through unchanged, and the worker routes forward requests directly to the Durable Object without creating a proxy.

valscion:
Nice! What's the story on network being slow? Do we deal with pending state changes in some separate state... or?

peterp:
> What's the story on network being slow?

It's immediate! When calling `setState` it will first update locally (in the browser), you can think of that as an optimistic update, and then it'll fire off a message to the server to sync that. I could probably add some additional information about the status of the sync.

valscion:
Thanks for the answer!

Then next question: How would an error be handled in the component when it is the first one to try to publish a change to the external store? Are we able to somehow show a toast message or similar to say that change's won't be synced but the app still works as if it was offline?

I would suppose that there would be a need to subscribe to the state synchronization state manually for monitoring or error tracking needs... e.g. something like using the same key in a different hook to know if the remote state is still being synced (basically did the worker respond with some success state) or if the transport failed. Granted, not all usages would need this, but some might.

The reason I ask these is I'm trying to find if the architecture would not be a good fit for some use case or if it accidentally lacked some key features. So basically I'm reviewing the concept 😄

docs/architecture/realtimeStateErrorHandling.md
```diff
@@ -0,0 +1,178 @@
+# Realtime State Error Handling and Offline Queue
```
peterp: @valscion I've just added this - this is a planned feature.

sdk/src/runtime/imports/client.ts
```diff
@@ -2,11 +2,39 @@ import React from "react";
 import { ClientOnly } from "../client/client";
 import { memoizeOnId } from "../lib/memoizeOnId";
 
-// @ts-ignore
-import { useClientLookup } from "virtual:use-client-lookup.js";
```
peterp: @justinvdm Gonna dig a bit deeper into this before I merge it.

valscion:
Thanks for the explainer!

I think this diverges somewhat unnecessarily from how `useOptimistic` is done in native React: https://react.dev/reference/react/useOptimistic

```jsx
export default function App() {
  const [messages, setMessages] = useState([
    { text: "Hello there!", sending: false, key: 1 }
  ]);
  async function sendMessageAction(formData) {
    const sentMessage = await deliverMessage(formData.get("message"));
    startTransition(() => {
      setMessages((messages) => [{ text: sentMessage }, ...messages]);
    })
  }
  return <Thread messages={messages} sendMessageAction={sendMessageAction} />;
}
```

Being able to set `sending: true` state in the frontend like the `useOptimistic` does with the optional second function in the example would be powerful:

```jsx
function Thread({ messages, sendMessageAction }) {
  const formRef = useRef();
  function formAction(formData) {
    addOptimisticMessage(formData.get("message"));
    formRef.current.reset();
    startTransition(async () => {
      await sendMessageAction(formData);
    });
  }

// ...

  const [optimisticMessages, addOptimisticMessage] = useOptimistic(
    messages,
    (state, newMessage) => [
      {
        text: newMessage,
        sending: true
      },
      ...state,
    ]
  );
```

It would make sense for the `useSyncState` to be modeled after `useOptimistic` over `useState` is what I think.

Notice how the callsite of `addOptimisticMessage` doesn't need to understand about `sending` flag at all but the `useOptimistic` second argument does the metadata addition instead.

peterp:
@valscion Sorry, I lost track of this conversation! Still trying to wrap my head around `useOptimistic`! I have not actually use this - and I'm trying to wrap my head around the benefits;

so essentially it's "state", but with a second function that allows for the definition of the optimistic update?

valscion:
> so essentially it's "state", but with a second function that allows for the definition of the optimistic update?

Yeah, it is "state" but indeed one where the second function can modify the `setState` calls to "decorate" the incoming state to also contain optimistic update related data, like converting from `string` type to `{ text: string, sending: boolean }` object so that UI can still keep on doing `setState(formData.get("someField"))` and not worry about the underlying data object being an object.

This moves the optimistic value wrapping over to the `useOptimistic` definition so the places where `setState` is being called don't have to worry about the value actually not yet being persisted.


> @valscion Sorry, I lost track of this conversation! 

Yeah I don't really know where would be a better place for this kind of async discussion, though? There's no RFC issue for this topic so that's why I continued over here 😄

PR #874, opened 2025-11-09
docs: Documents How To Send and Receive using Cloudflare Email Workers

dthyresson:
RedwoodSDK integrates with [Cloudflare Email Routing](https://developers.cloudflare.com/email-routing/email-workers/) to send, receive, reply, and forward email directly from your Worker.
This guide walks through the required configuration for production deployments, how to exercise the email handler locally,
and provides a reference worker illustrating the common patterns.

This PR documents how to implement and also test in local dev.

Note: The key change to support receiving emails is in how the default export works. 

It has to `export default class DefaultWorker extends WorkerEntrypoint<Env> {` and implement handlers for email and fetch with the app default.

Also: I did not change anything in https://docs.rwsdk.com/guides/email/sending-email/ for the Resend guide, but perhaps rename the guide to sending with Resend instead? --- yes, for generao=l send

Important:

I just re-read the Email Worker docs:

> destination_address: When you define the destination_address attribute, you create a targeted binding. This means you > can only send emails to the chosen email address. 
 
and 

> To prevent spam, email rules do not become active until after the destination address has been verified. Cloudflare 
> sends a verification email to destination addresses specified in Custom addresses. You have to select Verify email 
> address in that email to activate a destination address.

So sending may not be as useful as I thought for general sending -- but it might allow replies to one and if you know you send emails only to a handful of people.

** I need to update to reflect send destination limits vs reply to. And if general send, then use something like Resend.
I may need to update the docs here

EDIT:

See note below. I have revised to 1) say how current email worker must send to verified address 2) reply to is ok 3) mention upcoming Cloudflare Email Service that will allow greater transactional email support.

### DIFF
--- a/docs/src/content/docs/core/email.mdx
+++ b/docs/src/content/docs/core/email.mdx
@@ -3,9 +3,11 @@ title: Email
 description: Send and receive emails with Cloudflare Email
 ---
 
-RedwoodSDK integrates with [Cloudflare Email Routing](https://developers.cloudflare.com/email-routing/email-workers/) to send, receive, reply, and forward email directly from your Worker.
-This guide walks through the required configuration for production deployments, how to exercise the email handler locally,
-and provides a reference worker illustrating the common patterns.
+RedwoodSDK integrates with [Cloudflare Email Workers](https://developers.cloudflare.com/email-routing/email-workers/) so your application can send transactional messages, receive inbound mail, and reply in the same Worker runtime.
+
+Production deliveries currently require recipients to be verified through Cloudflare Email Routing, but Cloudflare’s forthcoming [Email Service beta](https://blog.cloudflare.com/email-service/) expands reach to general addresses.
+
+This guide walks through the configuration steps, highlights important sending considerations, and demonstrates common patterns for end-to-end email workflows.
 
 ## Implementing Email Handling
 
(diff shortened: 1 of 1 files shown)


peterp:
The ability to send email anywhere is coming: https://blog.cloudflare.com/email-service/

dthyresson:
> The ability to send email anywhere is coming: https://blog.cloudflare.com/email-service/

@peterp: 
Excellent! Ok, I updated the document to reflect the current state of Email Routing and Workers as well as document the forthcoming Email Service that can be used for general transactional emails.

Let me know if think these docs are still useful. It should be quick to update the sending email with the new service when out of beta -- worker looks like it would be same, I bet there is just more CF configuration and settings for spam/dmarc and such.

dthyresson:
I do have to admit, I am not sure what this:

```
/**
 * This is the default worker entrypoint for the Worker.
 * It extends the WorkerEntrypoint class and implements the email and fetch handlers.
 */
// It extends the WorkerEntrypoint class and implements the email and fetch handlers.
export default class DefaultWorker extends WorkerEntrypoint<Env> {
```

means, but I needed it otherwise when posting to the email handler in dev, I got an error saying the RPC needed to be a `WorkerEntrypoint` (or something like that)

See: https://developers.cloudflare.com/workers/runtime-apis/bindings/service-bindings/rpc/

justinvdm:
Thanks @dthyresson!

> > The ability to send email anywhere is coming: https://blog.cloudflare.com/email-service/
> 
> @peterp: Excellent! Ok, I updated the document to reflect the current state of Email Routing and Workers as well as document the forthcoming Email Service that can be used for general transactional emails.
> 
> Let me know if think these docs are still useful. It should be quick to update the sending email with the new service when out of beta -- worker looks like it would be same, I bet there is just more CF configuration and settings for spam/dmarc and such.

@peterp what's your take, should we merge and then revise once general email sending is around, or rather close and wait until its around?

peterp:
I think we should merge this and then create a new core document on email.

PR #875, opened 2025-11-09
Typescript inference for routes.

### LINKED ISSUE #311: Avoid repetition for routes+links
From https://discord.com/channels/679514959968993311/1307974274145062912/1361442272289493002

> Adding links to the link file seemed tedious. You essentially have to define routes twice. I understand that something you return a regular response from you might not want as a link but wondering if links could be auto generated for you.

peterp:
Each `defineApp` call now returns an object that carries the original route array on a private `__rwRoutes` field. TypeScript utility types walk that structure, inspecting nested `render`, `layout`, and `prefix` calls to assemble the union of normalized path strings.

### Link Factory

Expose `linkFor<App>()` and `createLinks<App>()`:

- `linkFor<App>()` returns the typed helper using a type-only reference to the app definition.
- `createLinks<App>()` is an alias for symmetry with the legacy API and remains available for users who prefer the naming.
- `defineLinks(routes)` stays as the manual escape hatch for tests that do not reference the app type.

Both helpers share a compile-time type that maps each route path to its expected params by parsing segments with template-literal types.

### Runtime Behaviour

The runtime interpolation mirrors the previous `defineLinks` implementation but derives its parameter validation directly from the provided argument. No manifests are required and the helper encodes path segments and wildcards before returning the final string.

### Client and Server Usage

Shared modules import the helper with a type-only reference to the worker:

```ts
import type App from "../../worker";
import { linkFor } from "rwsdk/router";

export const link = linkFor<App>();
```

Because the helper is created from types, client bundles do not pull in the worker implementation.

Closes #311

### DIFF
--- a/docs/architecture/autoGeneratedLinkHelpers.md
+++ b/docs/architecture/autoGeneratedLinkHelpers.md
@@ -29,13 +29,14 @@ The runtime interpolation mirrors the previous `defineLinks` implementation but 
 Shared modules import the helper with a type-only reference to the worker:
 
 ```ts
-import type App from "../../worker";
 import { linkFor } from "rwsdk/router";
+
+type App = typeof import("../../worker").default;
 
 export const link = linkFor<App>();
 ```
 
-Because the helper is created from types, client bundles do not pull in the worker implementation.
+Because the helper is created from types, client bundles do not pull in the worker implementation. The `typeof import("../../worker").default` alias keeps the reference purely at the type level while still giving `linkFor` access to the inferred routes.
 
 ## Tasks
 
@@ -43,4 +44,3 @@ Because the helper is created from types, client bundles do not pull in the work
 - Update the runtime link helper to operate without manifest inputs while still enforcing parameter usage.
 - Document the `linkFor` workflow and adjust starter/playground examples to rely on type-only imports.
 - Add unit tests for runtime interpolation and TypeScript tests (future) that exercise inference scenarios.
-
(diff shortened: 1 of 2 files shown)


justinvdm:
This is so cool!

justinvdm:
Going to review this tomorrow so I can give it a proper look, hope that is alright, let me know if you need this sooner.

peterp:
@justinvdm Let me just once over this as well! The majority of this code was vibe coded. I would actually prefer to add a performance benchmark before we merge this. I want to ensure that we're not adding a regression.

peterp:
@justinvdm Curious if you have any recommendations for bench-marking the router?

justinvdm:
> @justinvdm Curious if you have any recommendations for bench-marking the router?

Its a good question, quite tricky, depends what we wanting to get out of it. I'm thinking:
* Benchmarking router computation-time on its own might be too fast (as in, its not doing computationally intensive enough work) to warrant its own isolated js-only measuring
* Unless we wanted to do benchmarking for perf regressions? For this PR specifically I'm not sure if it changed much computational complexity-wise. But ya more generally might good, but then the value would be more relative comparison (% slower/faster)
* I think **what might be more interesting** in knowing up front, and perhaps also over time, would be real world benchmarks (actual deployed app). But then likely more meaningful to be looking from end to end what the perf is, rather than router alone. I think that'd give us numbers we can reason about better, use for identifying perf regressions over time, and also use as starting point for identifying bottlenecks in the entire process end to end

So bottom line, my gut feeling is: we'd get more bang for buck from something like a deploy of a sample application and running measurements on request and interaction metrics. There's definitely more to it that we'd need to think through (e.g. CF-related nuances like warm  vs cold workers and DOs), but maybe we think more about that when we get to it.

PR #877, opened 2025-11-10
docs: Updates to Cloudflare Email handling documentation

dthyresson:
Having now actually deployed a RWSK app to Cloudflare, I learned a new things:

- By default, the worker is not an email worker. You need to extend the `WorkerEntrypoint` class and implement the `email` handler to make it an email worker.
- By extending the `WorkerEntrypoint` class, you are telling Cloudflare that the worker is also email worker and can be selected to route inbound emails to.
- The `message` parameter is an `ForwardableEmailMessage` object that contains the inbound email message.
- The `In-Reply-To` header is required to reply to the inbound email.

This PR adds this information and corrects some example code to reply to an inbound email.

I do have both sending as well as replying to an inbound email working in a deployed app.

### DIFF
--- a/docs/src/content/docs/core/email.mdx
+++ b/docs/src/content/docs/core/email.mdx
@@ -47,6 +47,10 @@ export it in the default export of the worker.
 
 This is a change to how `defineApp` is often used in the worker.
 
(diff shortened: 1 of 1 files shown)


justinvdm:
Appreciate you keeping this updated and accurate with what you learned, thank you @dthyresson!

PR #878, opened 2025-11-11
fix(db): Infer column nullability correctly for rwsdk/db

kcc989:
# Add Support for Nullable Columns in Type System

## Problem

I was trying to add a table to my project

```
export const centralMigrations = {
  '001_taxonomy_tables': {
    async up(db) {
      return [
        await db.schema
          .createTable('ingredients')
          .addColumn('id', 'text', (col) => col.primaryKey())
          .addColumn('name', 'text', (col) => col.notNull())
          .addColumn('category', 'text')
          .addColumn('created_at', 'text', (col) => col.notNull())
          .addColumn('updated_at', 'text', (col) => col.notNull())
          .execute()
       ]
    },

    async down(db) {
      await db.schema.dropTable('ingredients').ifExists().execute();
    },
  },
} satisfies Migrations;
```

When trying to insert into this table I got errors with TypeScript because it wouldn't accept `null` for `category`. The SDK was missing support for null columns in the type system. In SQL, the standard is to have columns be nullable by default and it becomes not null based on the constraints added to the column. 

## Solution
This PR adds nullable column support to the SDK db's type inference system. The main changes are:
* Added a `TNullable` type parameter to `ColumnDefinitionBuilder`, `addColumn`, and `modifyColumn` operations that defaults to `true`. This gives us default nullability. 
* Specific column constraints now correctly make a column not null. `primaryKey()`, `notNull(), and `defaultTo()` will make a column not null. 
* Other constraints like `unique()` allow the column to stay null

This works across creating, adding, and modifying a column. 

## Testing

Added additional coverage to the `typetest.ts` files to handle more scenarios and properly check for null across constraint functions. Also updated existing test cases to either specific not null or allow nullable values. 

***

### Additions: Support for Generated Columns on Insert/Update (justinvdm)

Building on @kcc989's awesome work, I extended the type system to correctly handle generated columns (`defaultTo`, `autoIncrement`, etc.) for insert and update operations.

**Problem**

While the previous changes correctly handled nullability for `select` types, the developer experience for `insert` and `update` operations was still incorrect. The type system required users to provide values for columns that the database would generate automatically. For example, an `autoIncrement` `id` or a `status` with `defaultTo('active')` were required fields on insert, which is not how the database behaves.

**Solution**

The core challenge was that Kysely requires schema information (using its `Generated<T>` type) to know which columns are generated, but our user-facing `Database` type must return plain, primitive types to avoid being a breaking change for existing user code.

This was solved by introducing an internal-only mechanism:

1.  The type inference engine was enhanced to track more metadata about each column (`isNullable`, `hasDefault`, `isAutoIncrement`).
2.  The `Database<T>` utility now produces a special `__kyselySchema` property alongside the plain, user-facing table types. This internal property contains a version of the schema where generated columns are correctly wrapped with Kysely's `Generated<T>` type.
3.  The `createDb()` function was updated to detect and use this `__kyselySchema`, providing Kysely with the metadata it needs under the hood.

The result is that the user-facing `Database` type remains clean and unchanged, preserving backward compatibility, while `insertInto` and `updateTable` operations now have the correct, more permissive types.

**Testing**

*   Added new type tests to verify that columns with `autoIncrement` and `defaultTo` are optional on insert, and that partial updates are now correctly typed.
*   Updated existing type tests to use a new `OmitInternals` helper, allowing them to ignore the new `__kyselySchema` property and continue to validate the public-facing schema.

***

### ⚠️ Breaking Change: Database Type Inference for Nullable Columns is Now Correct

This release corrects a fundamental issue in the `rwsdk/db` type inference system where columns were incorrectly inferred as non-nullable by default. The system now correctly aligns with SQL standards, where columns are nullable unless explicitly constrained.

**What is the change?**

*   **Previous Behavior:** All columns were typed as non-nullable (e.g., `string`, `number`) unless you were using a version that had partial fixes.
*   **New Behavior:** Columns are now correctly inferred as nullable by default (e.g., `string | null`, `number | null`). A column will only be non-nullable if it has a `.notNull()`, `.primaryKey()`, or `.defaultTo()` constraint in its migration definition.

**Why is this a breaking change?**

This change fixes a bug in the type system. If your application code was written relying on the previously incorrect non-nullable types, you may now see new TypeScript errors.

These errors are not new bugs. They are surfacing legitimate type mismatches that were previously hidden by the incorrect inference. Your code may have been vulnerable to runtime errors if the database returned a `null` value that the type system did not account for.

**How to update your code**

You will need to update any code that interacts with nullable columns to correctly handle the possibility of `null` values. The TypeScript errors will guide you to the exact locations.

For example, if you have a `users` table where the `bio` column is nullable:

```typescript
// src/db/migrations.ts
// ...
  .createTable("users")
  .addColumn("id", "text", (col) => col.primaryKey())
  .addColumn("bio", "text") // This is nullable by default
// ...
```

Your application code might now show a type error:

```typescript
import { db } from "@/db";

const user = await db.selectFrom("users").selectAll().executeTakeFirst();

// 🚨 ERROR: 'user.bio' is possibly 'null'.
// Type 'string | null' is not assignable to type 'string'.
const bioLength: number = user.bio.length;
```

**To fix this, add a null check before accessing the property:**

```typescript
import { db } from "@/db";

const user = await db.selectFrom("users").selectAll().executeTakeFirst();

// ✅ Correct: Handle the null case
const bioLength: number = user.bio ? user.bio.length : 0;
```

By making these adjustments, your application will be more robust and correctly typed against the actual schema of your database.

### DIFF
--- a/docs/architecture/databaseTypeInference.md
+++ b/docs/architecture/databaseTypeInference.md
@@ -1,0 +1,82 @@
+# Database Type Inference
+
+This document outlines the strategy used to provide strong, zero-config type safety for `rwsdk/db`. The system infers the complete database schema by statically analyzing the user's migration files at the type level, providing developers with full type safety and autocompletion without a code generation step.
+
+## The Challenge: Avoiding Code Generation
+
+Many database tools rely on a code generation step to provide type safety. While effective, this approach introduces several points of friction into the development process:
+
+1.  **An Extra Step**: It requires the developer to run a command every time the schema changes, which is an easy step to forget and a common source of type-related errors.
+2.  **Fragile Tooling**: The code generation process itself becomes another moving part that can fail. It can be sensitive to configuration (like file paths), and upgrades to the underlying tools can change the generated output, requiring workarounds.
+3.  **Redundant Information**: Fundamentally, the database schema is already fully defined in the migration files. A code generation step is a workaround for a tool's inability to read this information directly.
+
+The goal of `rwsdk/db`'s type system is to eliminate these problems by inferring the schema directly from the code that is already the source of truth: the migrations.
+
+This presents two primary technical challenges: how to derive a final schema from a sequence of migrations, and how to represent that schema for different database operations, all at the type level.
+
+### 1. Inferring a Schema from a Sequence of Changes
+
+The database schema is not defined in a single file; it is the result of a sequence of migration files. The type system needed a way to compute the *final* state of the schema by simulating this entire migration process at compile time, without ever connecting to a database or executing runtime code.
+
+### 2. Schemas for Different Operational Contexts
+
+A single database table does not have a single, universal type. Its type signature changes depending on the operation being performed:
+
+*   **For `select` operations**: When reading data, all columns are present as they exist in the database.
+*   **For `insert` operations**: When creating a new record, columns that the database generates automatically (e.g., via `autoIncrement` or `defaultTo`) must be optional.
+*   **For `update` operations**: When updating a record, all columns must be optional, as any subset of fields can be changed.
+
+The type system needed to be able to generate and provide the correct type signature for each of these contexts.
+
+### 3. Integrating with Kysely's Type System
+
+The underlying database client, Kysely, has its own type system. To enable its `insert` and `update` type inference, it requires schema definitions to explicitly mark generated columns with a special `Generated<T>` utility type (e.g., `{ id: Generated<number> }`). This system is not designed to be extended by an external, dynamic type-inference engine like ours.
+
+This created a challenge: our user-facing `Database` type must return plain, primitive types to ensure a simple developer experience (e.g., `type User = DB['users']`), but Kysely needed the `Generated<T>` wrappers to function correctly.
+
+## The Solution
+
+The solution is a multi-stage, purely type-level process that simulates the migrations to build a metadata-rich schema, which is then used to generate the different schema signatures required by both the developer and the underlying Kysely client.
+
+### 1. Sequential Type-Level Migration Processing
+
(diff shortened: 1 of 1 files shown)


kcc989:
This doesn't help with `defaultTo` columns needing to be specified during insert. I couldn't figure out how to do that without some large changes to the type system. I think that would require having separate insert and select types.

justinvdm:
Thank you for this @kcc989, this is awesome. Just taking some time to grok the change, its impact (e.g. it reveals types were technically broken before) and what's left to do re the `defaultTo`. Let me know if you need a test release for this change in the meantime.

justinvdm:
Thank you again for tackling this. @kcc989. You're right, the type system was fundamentally broken for nullable columns, and this PR correctly addresses that.

Even though it's an important fix, this will be a breaking change. I've also been looking at the remaining issues you flagged with `defaultTo`. You correctly pointed out that the type inference approach we currently have is going to need quite a bit more reworking to solve that. So while we could likely ship those fixes later in a non-breaking way, I think it makes more sense to do a single release for all these changes in behaviour.

I'm happy to take it from here to get the rest of it done. In the meantime, if you want to use your fix right away, I've published a test release: `1.0.0-beta.27-test.20251111115809`.

Really appreciate you bringing solving this all so far already! 🙇

kcc989:
No problem! I'm happy to help. 

And thanks for shipping the test release, I really appreciate it.

justinvdm:
Ok I now have a working solution that includes `defaultsTo` columns no longer needing to be provided during insert.

I'll be away most of the time until next week, so I'll do a proper release for this when I'm back. Until then, here's a test release that includes the `defaultsTo` fix: `1.0.0-beta.27-test.20251112194755`

kcc989:
Thanks @justinvdm. Did you push that change? I'd just be curious to see the code

EDIT: I see it in the test release changes

PR #879, opened 2025-11-12
Avoid rendering current docs hover color with background color

valscion:
Otherwise hovering or focusing on the current docs page will turn the text color to be the same color as the background color is — essentially making the text disappear.

Here's screenshots from Firefox:

## Baseline (no hover or focus visible)

image

### Hover

image

### Focus visible

image

### Hover

image

### Focus visible

image

Essentially nothing changes. This matches the Starlight CSS https://starlight.astro.build/getting-started/ as they also don't change the text color or background color of the active page on hover or focus visible.

### DIFF
--- a/docs/src/styles/custom.css
+++ b/docs/src/styles/custom.css
@@ -11,8 +11,8 @@
   color: #0f172a !important;
 }
 
-.sidebar-content a:hover,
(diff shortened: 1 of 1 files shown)


peterp:
Thank you!

PR #885, opened 2025-11-16
fix: Ensure hoisted meta tags are correctly placed in document head

### LINKED ISSUE #872: Title and Meta tags
Hi there! 👋

First of all — fantastic work on this project! It fits my needs incredibly well. 🙌

I had a quick question about the behavior described in the [Frontend Metadata guide](https://docs.rwsdk.com/guides/frontend/metadata/).
From the docs, it looks like the `Title` and `Meta` components should “just work” out of the box — but currently, they seem to only update the title and meta tags **in the browser**, not during **server-side rendering** / initial rendering.

Is that the intended behavior?
I can manually set the title via `ctx` and pass it up to the server, but if the goal was for metadata to be handled automatically on the server as well, maybe the guide could mention that distinction, since it's a bit important for SEO?

For reference, here’s what the rendered HTML looks like right now:

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charSet="utf-8"/>
  <meta name="viewport" content="width=device-width, initial-scale=1"/>
  <title>The title</title>
  <link rel="modulepreload" href="/src/client.tsx"/>
</head>
<body>
  <div id="root">
    <!-- omitted script tag -->
    <div id="hydrate-root">
      <title>The title specified in home</title>
    </div>
  </div>
</body>
</html>
```

justinvdm:
#### Problem

During server-side rendering, React's hoisted tags (e.g., `<title>`, `<meta>`) were incorrectly rendered in the `<body>` instead of the document `<head>`. This occurred because the framework's two-stream rendering architecture sent the document shell, including the `<head>`, before the application stream containing the hoisted tags was processed. While React correctly hoisted the tags to the beginning of the application stream, they arrived too late to be placed in the document head.

#### Solution

This change introduces a stream-splitting utility that pre-processes the application's HTML stream. The function separates the stream into two distinct streams: one containing the hoisted tags from the beginning of the stream, and a second containing the remainder of the application body.

The main stream-stitching logic is updated to read all hoisted tags first, then process the outer document stream. When the `</head>` closing tag is detected, the hoisted tags are injected immediately before it, ensuring they appear at the bottom of the `<head>` section, after any existing head content. The doctype declaration remains the first element in the output, maintaining valid HTML structure. The rest of the stitching logic proceeds unchanged, interleaving the document and application body streams as before.

Fixes #872

### DIFF
--- a/sdk/src/runtime/lib/stitchDocumentAndAppStreams.ts
+++ b/sdk/src/runtime/lib/stitchDocumentAndAppStreams.ts
@@ -240,10 +240,12 @@ export function stitchDocumentAndAppStreams(
         // Continuously read from the hoisted tags stream and buffer the
         // content. Once the stream is finished, transition to the next phase.
         const { done, value } = await hoistedTagsReader.read();
+        // When the stream is done, we're ready to process the document head.
         if (done) {
           hoistedTagsReady = true;
           phase = "outer-head";
         } else {
+          // Otherwise, keep appending to the buffer.
           hoistedTagsBuffer += decoder.decode(value, { stream: true });
         }
       } else if (phase === "outer-head") {
@@ -252,7 +254,9 @@ export function stitchDocumentAndAppStreams(
         // `startMarker` to know when to start injecting the app shell. Once
         // the marker is found, transition to the next phase.
         const { done, value } = await outerReader.read();
+        // Handle the case where the document stream ends.
         if (done) {
+          // If there's content left in the buffer, process it for markers.
           if (buffer) {
             const headCloseIndex = buffer.indexOf("</head>");
             if (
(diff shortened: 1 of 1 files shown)


Tobbe:
> The main stream-stitching logic is updated to first exhaust and prepend the hoisted tags stream to the response, ensuring they appear before the `<!DOCTYPE html>` declaration.

Are you putting the `<title>` etc tags before the doctype? Are you sure that's the correct thing to do? I thought the doctype had to be the very first thing arriving to the browser so it doesn't enter quirks mode.

justinvdm:
> > The main stream-stitching logic is updated to first exhaust and prepend the hoisted tags stream to the response, ensuring they appear before the `<!DOCTYPE html>` declaration.
> 
> Are you putting the `<title>` etc tags before the doctype? Are you sure that's the correct thing to do? I thought the doctype had to be the very first thing arriving to the browser so it doesn't enter quirks mode.

Great catch, you're 100% right. Not only this - we should be putting them in the `<head>`, don't know how I missed this. Will fix, thank you!

PR #884, opened 2025-11-16
Playground for use-sync-state

peterp:
Regarding the big diff: I also removed the cloudflare generated types from the playgrounds.

@justinvdm I'm having a tough time getting capnweb working in our SSR environment. Capnweb has two exports: "workerd" as well as "default," but the "workerd" version is getting pulled into SSR which has: `import { env } from "cloudflare:workers"`

To reproduce; try to `pnpm build` or `pnpm dev` the `use-synced-state` playground.

I've tried fixing this on the configuration level, and adding `"use client"` to the `use-synced-state` code where it's relevant, but my Vite foo is lacking.

PR #891, opened 2025-11-19
Fix useSyncedState exports

kcc989:
This should allow use of the hook introduce in #871. 

I wasn't able to import the synced state hook or DO like shown in [the docs](https://docs.rwsdk.com/core/usesyncedstate/). After doing a little digging it seems like the exports were messed up in package.json. The hook wasn't being exported at all and the DO was referencing nonexistant files. I built locally and think this should be fine. 

cc: @peterp

### DIFF
--- a/sdk/package.json
+++ b/sdk/package.json
@@ -105,13 +105,17 @@
       "types": "./dist/runtime/lib/realtime/durableObject.d.ts",
       "default": "./dist/runtime/lib/realtime/durableObject.js"
     },
+    "./use-sync-state": {
+      "types": "./dist/use-synced-state/useSyncedState.d.ts",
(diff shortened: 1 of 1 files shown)


peterp:
Thanks @kcc989 I'm taking a look at this right now. I still need to test using synced state; There's a playground that is failing, but I'll get to this as soon as possible.

kcc989:
No problem. I looked through the playground PR after I posted this. I know its experimental. I just wanted to mess around a bit 😆

peterp:
Just merged the new stuff, will release a test version shortly.

peterp:
Closing due to this: https://github.com/redwoodjs/sdk/pull/884

Thanks though!

PR #898, opened 2025-11-21
Fix type hover for routers triggering wrong Markdown formatting

valscion:
The type hover signature in e.g. VS Code treats the text as markdown and thus the `*, /api/*` part is considered to be `<em>, api/</em>` unless the `*` is escaped.

I've verified these changes by tweaking the contents of `node_modules/rwsdk/dist/runtime/lib/router.d.ts` to incorporate these same changes.

This is basically me following up with what https://github.com/redwoodjs/sdk/pull/866 changes actually look like in the editor.

## Before

SCR-20251121-mwby

## After

SCR-20251121-mvzr


peterp:
Thanks!

PR #899, opened 2025-11-22
feat: Update `renderToString` options to include all `renderToStream` options

gching:
## Problem
* `renderToString` does not take in all options for `renderToStream`, in this case for now, `ssr`

## Solution
* Use `RenderToStreamOptions` instead and omit `onError`


justinvdm:
Thanks @gching!

justinvdm:
Released in 1.0.0-beta.32

PR #903, opened 2025-11-24
fix: Fix font loading 404s in production

justinvdm:
## Problem

CSS `@import` for fonts (like `@fontsource/figtree`) works great in dev but falls over in production. The build puts the font files into `dist/worker/assets` instead of `dist/client/assets`, so the browser gets a 404 when trying to load them.

We had a similar issue with CSS files before, which is why we have the `moveStaticAssetsPlugin` to shuffle `.css` files over to the client build. But it was too picky and only looked for CSS files.

## Solution

I've updated `moveStaticAssetsPlugin` to be less picky. It now moves **all** static assets (except for `.js` and `.map` files) from the worker build to the client build.

This means any asset referenced by CSS—fonts, images, whatever—that ends up in the worker bundle will now get moved to the client distribution where it belongs.

## Context & Future Investigation

The `moveStaticAssetsPlugin` was originally a bit of a patch to handle CSS file generation during the worker build. While this change gets fonts working, it points to a bigger architectural question about how we handle assets across the split `worker`+`client` environments.

In a perfect world, we wouldn't need to manually shuffle assets around like this. It's worth investigating if we can get the deploy step to look at both `dist/worker` and `dist/client` for assets.

I do remember trying to get all the assets to output to a single directory, but that route proved to be tricky in practice. I think it was partly the challenge of one env clearing the other env's assets (among other challenges) thought I could be misremembering.

### DIFF
--- a/.notes/justin/worklogs/2025-11-24-fontsource-css-imports-repro.md
+++ b/.notes/justin/worklogs/2025-11-24-fontsource-css-imports-repro.md
@@ -38,3 +38,27 @@ Updated `moveStaticAssetsPlugin` to also move font files (`.woff`, `.woff2`, `.t
 3. Move all matching files to `dist/client/assets/`
 
 This ensures font files referenced in CSS are accessible to the browser in production builds.
+
+---
+
+Fix font loading 404s in production by moving all static assets
+
+## Problem
+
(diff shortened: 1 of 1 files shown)


Zn4rK:
I haven't looked too far into RedwoodSDK's code base just yet to say if this is relevant or not, but Remix / React Router moves all assets emitted by ssr into the client build;
https://github.com/remix-run/react-router/blob/a947b0346eba3abd0ca771bc19640ef1f20e271e/packages/react-router-dev/vite/plugin.ts#L1728 

So, it looks like it behaves the same way as this code :)

Issue #865, opened 2025-11-05
Feature: Allow custom type serializers/parsers in rwsdk/db


justinvdm:
### The Problem

Currently, `rwsdk/db` returns raw data from SQLite, meaning types like dates must be handled as primitives (`string` or `number`). This requires developers to manually serialize values before writing to the database and parse them after reading.

For example, when working with `Date` objects, a developer has to add boilerplate for conversion:

```typescript
// Manually serializing a Date object to an ISO string before inserting
await db.insertInto("todos").values({
  // ...
  createdAt: new Date().toISOString(),
}).execute();

// Manually parsing the ISO string back into a Date object after fetching
const todoFromDb = await db.selectFrom("todos").selectAll().executeTakeFirst();
const createdAtDate = new Date(todoFromDb.createdAt);
```

This process is repetitive and can lead to inconsistencies.

### The Proposed Solution

Introduce a mechanism that allows users to provide their own type serializers and parsers. This would let developers work with rich types like `Date` objects directly in their application code, while the database client handles the transformation to and from SQLite-compatible types automatically.

This pattern is common in other database clients and query builders (e.g., Kysely's `DataTypeTransformer` or Drizzle's `customType`).

An example of what the user-facing API could look like:

```typescript
// An example of a type converter configuration
const typeConverters = {
  Date: {
    // Invoked when reading from the database
    parse: (value: string | number | null) => (value ? new Date(value) : null),
    // Invoked when writing to the database
    serialize: (value: Date | null) => (value ? value.toISOString() : null),
  },
};

// The converters could be passed as an option to `createDb`
export const db = createDb<AppDatabase>(
  env.APP_DURABLE_OBJECT,
  "todo-database",
  { typeConverters }
);

// With this configuration, developers could work directly with Date objects:
const todo = await db.selectFrom("todos").selectAll().executeTakeFirst();
console.log(todo.createdAt.getFullYear()); // `createdAt` is a Date object

await db.insertInto("todos").values({
  // ...
  createdAt: new Date(), // The value is automatically serialized
}).execute();
```

### Alternatives Considered

A default, built-in handler for `Date` objects was considered. However, this approach is less flexible. Different projects may have different requirements for serialization (e.g., storing dates as Unix timestamps vs. ISO strings, or handling timezones). A customizable system is more adaptable and avoids making assumptions about user needs.

justinvdm:
[...]

PR #904, opened 2025-11-25
feat(db): Add support for custom type converters

Bittu-the-coder:
## Problem
Currently, `rwsdk/db` returns raw data from SQLite, restricting developers to primitive types (string or number). This necessitates manual serialization of complex types like `Date` before writing to the database and parsing them back after reading, leading to repetitive boilerplate code and potential inconsistencies.

## Solution
This PR introduces a flexible mechanism for custom type converters, allowing developers to work with rich types directly in their application code.

### Key Changes:
- **New `typeConverters` option**: Added to [createDb](cci:1://file:///c:/Users/bittu/Documents/oss/sdk/sdk/src/runtime/lib/db/createDb.ts:9:0-33:1) to accept a configuration object for custom types.
- **[DOWorkerDialect](cci:2://file:///c:/Users/bittu/Documents/oss/sdk/sdk/src/runtime/lib/db/DOWorkerDialect.ts:21:0-43:1) update**:
    - Implemented [processParameters](cci:1://file:///c:/Users/bittu/Documents/oss/sdk/sdk/src/runtime/lib/db/DOWorkerDialect.ts:115:2-135:3) to automatically serialize objects (e.g., `Date` -> ISO string) before query execution.
    - Implemented [processRows](cci:1://file:///c:/Users/bittu/Documents/oss/sdk/sdk/src/runtime/lib/db/DOWorkerDialect.ts:137:2-164:3) to automatically parse raw database values back into rich objects based on column names or custom matchers.
- **Type Definitions**: Added `TypeConverter` and `TypeConverters` interfaces in `typeConverters.ts`.

## Usage Example

```typescript
const typeConverters = {
  Date: {
    // Match columns ending in 'At' (e.g., createdAt)
    match: (col) => col.endsWith("At"),
    // Parse string from DB to Date object
    parse: (value) => (value ? new Date(value) : null),
    // Serialize Date object to ISO string for DB
    serialize: (value) => (value ? value.toISOString() : null),
  },
};

export const db = createDb(env.DB, "main", { typeConverters });

// Now works directly with Date objects:
await db.insertInto("todos").values({ createdAt: new Date() }).execute();

### CODE REVIEW

**Comment by justinvdm** [comment]
This is awesome, thank you @Bittu-the-coder! Checking now

Fixes #865

File: sdk/src/runtime/lib/db/DOWorkerDialect.ts [hunk 1]
```diff
@@ -92,13 +100,68 @@ class DOWorkerConnection implements DatabaseConnection {
...
+
+    return rows.map((row) => {
+      const newRow = { ...row };
+      for (const key in newRow) {
+        const value = newRow[key];
+        // 1. Check for exact column name match
+        if (this.typeConverters![key]?.parse) {
+          newRow[key] = this.typeConverters![key].parse!(value, key);
+          continue;
+        }
```
justinvdm: I think it might be a bit too implicit that exact column names also count as matches, or might come as a surprise to some users.

What do you think of us relying on `match()` always? That way its explicit what matches and what doesn't.

sdk/src/runtime/lib/db/typeConverters.test.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    const mockExecuteQuery = vi.fn().mockResolvedValue({
+      rows: [
+        { id: 1, createdAt: "2023-01-01T00:00:00.000Z" },
+        { id: 2, createdAt: null },
+      ],
+    });
+
+    const dialect = new DOWorkerDialect({
+      kyselyExecuteQuery: mockExecuteQuery,
```
justinvdm: Clever using this as dependency injection for tests!

justinvdm:
Left some minor comments, looks solid, thank you again!

justinvdm:
Something I'm not sure how we handle is arrays of values, but I guess those can be types themselves. Regardless, something for us to figure out later once we've gotten a feel for this ^ feature.

sdk/src/runtime/lib/db/typeConverters.test.ts
Bittu-the-coder: Thanks, @justinvdm!

Bittu-the-coder:
I’ve updated the PR. Please review the changes.

justinvdm:
Hi @Bittu-the-coder,

First off, thank you for all the work you've put into this PR. It's a clean implementation and it perfectly addresses the issue I created.

I've been thinking more about this since then and wanted to share where my head is at. Please bear with me here - it's going to take a bit of explaining.

### My thinking

I've been trying to imagine how this feature would look in a growing project. The core challenge is that any automatic type conversion only has the column name to work with, since the table context from the original query is lost by the time we get the results.

For example, I could see a central config accumulating complex rules over time:

```typescript
// Example of a config in a growing project
const typeConverters = {
  Date: {
    match: (col) => {
      // Simple at first...
      if (col.endsWith('At')) return true;
      // ...then exceptions creep in
      if (col === 'last_login') return true;
      if (col === 'status_changed_on_date') return true;
      // This gets hard to maintain, especially with name collisions across tables.
      return false;
    },
    // ...
  },
};
```

This feels brittle - for example it can break with simple things like aliasing (`.select('createdAt as joined')`). It also becomes unwieldy to maintain as the schema grows: needing to add/remove/rename columns in this central config every time the database changes.

This has led me to believe that the better way to support type conversion in RedwoodSDK, would be to provide the type converters with context on the database structure - the relevant table and column name, possibly also type information. This would be ideal, but it's a huge architectural addition that we can't commit to supporting right now.

I know it was me that suggested this idea in my issue (#865) in the first place. I'm sorry for leading us down this same path that I'm now backpedalling on.

### Two Paths Forward

With all that said, I see two paths forward that could unblock you.

**1. A User-Land Kysely Plugin**

**You can already implement this today without any SDK changes.** When extending `SqliteDurableObject`, you can pass in a custom Kysely Plugin to handle serialization and parsing. This is Kysely's intended pattern for this.

```typescript
// In your app's Durable Object file
class MyDatabase extends SqliteDurableObject {
  constructor(state, env) {
    super(state, env, migrations, undefined, [
      new MyTypeConverterPlugin({
        'createdAt': (val) => new Date(val),
        'last_login': (val) => new Date(val),
      })
    ]);
  }
}
```

Granted, this has the same limitation as what I described above. The difference is it is in user-land, so if this is sufficient for your project's purpose, then that's great. We just can't provide this as a built-in solution because of these limitations.

**2. The "Ideal Solution" as an External Package**

The runtime introspection idea could be built as a standalone package. It's a big project, but it is the "holy grail" solution that aligns with our philosophy of using migrations as the source of truth for database structure.

```typescript
// (Heavily pseudocoded)
import { createTypeConverterPlugin } from 'rwsdk-db-converters';


// This factory executes migrations against a mock runner to collect schema,
// then uses that schema to provide full context to your converter functions.
const typeConverterPlugin = createTypeConverterPlugin(migrations, {
  // User provides the conversion logic, the plugin provides the context.
  serialize: (ctx) => {
    // ctx = { tableName: 'users', columnName: 'createdAt', value: Date(...) }
    if (ctx.tableName === 'users' && ctx.columnName === 'createdAt') {
      return (ctx.value as Date).toISOString();
    }
    return ctx.value;
  },
  
  parse: (ctx) => {
    // ctx = { tableName: 'users', columnName: 'createdAt', value: '2023-...' }
    if (ctx.tableName === 'users' && ctx.columnName === 'createdAt') {
      return new Date(ctx.value as string);
    }
    return ctx.value;
  }
});

// In your app's Durable Object file:
class MyDatabase extends SqliteDurableObject {
  constructor(state, env) {
    super(state, env, migrations, undefined, [typeConverterPlugin]);
  }
}
```

If you ever explored this, I'd be happy to feature it in our docs. Obviously no pressure, but I do think this is the better way forward, and what we'd be more likely to do if we committed to supporting it. Perhaps there's even an opportunity for us to bring the solution into core at a later stage if it matures (no promises, just leaving the door open).

---

I hope this all makes sense. What do you think?

Bittu-the-coder:
Thanks for the detailed explanation, @justinvdm — this makes complete sense.

I appreciate you taking the time to walk through the long-term implications. You're right that relying only on column names becomes brittle quickly, and without table/column context the SDK can’t reliably support this in a scalable way.

I'm totally fine with shifting direction.
If this approach is better kept as a user-land Kysely plugin, I’m happy to adapt my use case that way. And the introspection-based external package idea sounds interesting — I'd love to explore it at some point.

Let me know whether you'd prefer that we close this PR or keep it open in case it can be repurposed (e.g., as documentation or an example). I'm good with whatever path you think is best.

Thanks again for the thoughtful feedback!

justinvdm:
> Thanks for the detailed explanation, @justinvdm — this makes complete sense.
> 
> I appreciate you taking the time to walk through the long-term implications. You're right that relying only on column names becomes brittle quickly, and without table/column context the SDK can’t reliably support this in a scalable way.
> 
> I'm totally fine with shifting direction. If this approach is better kept as a user-land Kysely plugin, I’m happy to adapt my use case that way. And the introspection-based external package idea sounds interesting — I'd love to explore it at some point.
> 
> Let me know whether you'd prefer that we close this PR or keep it open in case it can be repurposed (e.g., as documentation or an example). I'm good with whatever path you think is best.
> 
> Thanks again for the thoughtful feedback!

Thanks so much for being so understanding @Bittu-the-coder! I really appreciate it.

I think the best path forward is for me to close this PR. For context, I'll link to our discussion here in the original issue before I close that as well.

If you do decide to explore either the user-land plugin or the more advanced introspection package and want to talk through the implementation or bounce some ideas around, please don't hesitate to reach out. We'd be happy to chat more. You can find us on Discord at https://discord.gg/redwoodjs.

Thanks again for your great work and for being so collaborative on this!

PR #906, opened 2025-11-28
fix: remove node_modules guard from entry branching + add playground example

Zn4rK:
This PR makes two improvements to RedwoodSDK:

1. **Removed the `node_modules` exclusion guard in the entry-resolution logic**
   The SDK previously had an `if` statement that stopped the branching logic from running whenever the resolved entry path contained `node_modules`. This prevented RedwoodSDK from handling entries that legitimately resolve to packages.

   I removed this `node_modules` check so that the branching logic always runs, regardless of where the entry is located. This makes the behavior consistent and fixes cases where code inside `node_modules` should still be treated as an entry point.

   *(Note: in the new playground example, the client entry is not in node_modules; it resides in the playground src due to a symlinking limitation.)*

2. **Added a new playground example**
   I added a minimal example to the `playground/` directory based on the existing “Hello World” example. This provides a simple, reproducible environment to test the updated entry-resolution behavior.


.gitignore
```diff
@@ -200,4 +200,7 @@ starters/standard/.pnpm-store
 .cursor/rules/user
 
 **/*.tsbuildinfo
-**/worker-configuration.d.ts
\ No newline at end of file
+**/worker-configuration.d.ts
+
+# Webstorm
+.idea/
```
Zn4rK: I can remove this if needed. I didn't find any obvious place for IDEs to place this under

justinvdm:
Thank you for working on this @Zn4rK! Appreciate you taking time on the E2E test - I know it was a can of worms.

Will keep an eye out for when CI is green, let me know if it'd help for me to take a look as well.

justinvdm:
Update for posterity:

We're skipping the e2e test for this playground example. Our e2e infrastructure runs tests in isolated directories with rwsdk as a tarball (not workspace-linked), which makes testing node_modules scenarios in playground examples difficult. The change works in real-world usage (as confirmed by @Zn4rK), and the e2e tests pass locally, but testing it in our playground setup on CI requires workarounds that aren't worth the rabbit hole to get working right now. We'll revisit e2e testing for monorepo/package scenarios when we have more of these use cases.

For context, see the discussion  on discord: https://discord.com/channels/679514959968993311/1443579628760334516/1445048899675885608

justinvdm:
Released in https://github.com/redwoodjs/sdk/releases/tag/v1.0.0-beta.34

Thank you @Zn4rK!

PR #920, opened 2025-12-03
Improve useSyncedState

peterp:
- [x]  Improve docs.
- [x] Validate that everything works


peterp:
Finally got the Playground E2E tests to work. At some stage I would love to improve the performance of those.

peterp:
https://developers.cloudflare.com/durable-objects/best-practices/rules-of-durable-objects/#do-not-use-a-single-durable-object-as-a-global-singleton

^ Take the advice in this to account. We currently use a single DO for all "states" interactions. According to this we should consider using a DO per namespace. So we'll abstract away the DO interface.

PR #921, opened 2025-12-04
fix: Avoid moving WASM assets

### LINKED ISSUE #913: WASM files not copied to worker assets after rwsdk beta.31
# WASM files not copied to worker assets directory

## Description

After upgrading to rwsdk beta.31+, deploying a project that uses WASM modules fails with:

```
Uncaught Error: No such module "assets/yoga-ZMNYPE6Z-CP4IUfLV-CP4IUfLV.wasm".
  imported from "index.js"
[code: 10021]
```

WASM files are placed in `dist/client/assets/` but not `dist/worker/assets/`, causing Wrangler to fail validation.

## Reproduction

1. Use rwsdk >= 1.0.0-beta.31
2. Install a package that uses WASM imports (e.g. `workers-og`, which depends on `yoga-wasm-web` and `@resvg/resvg-wasm`)
3. Run `yarn build && wrangler deploy`

## Root Cause

Introduced in commit `8bf62d98` (fix(build): Support external modules in SSR bundle #893).

This commit changed the SSR bundle wrapping to keep imports at the top level instead of wrapping them inside an IIFE. As a side effect, WASM imports like `import "./assets/yoga-....wasm"` are now preserved as ES module imports in the worker bundle, but the actual `.wasm` files are only output to `dist/client/assets/`, not `dist/worker/assets/`.

## Workaround

Copy WASM files from client to worker assets before deploying:

```json
{
  "scripts": {
    "deploy": "yarn build && cp dist/client/assets/*.wasm dist/worker/assets/ && wrangler deploy"
  }
}
```

## Expected Behavior

WASM files imported by worker code should be automatically copied to `dist/worker/assets/`.

justinvdm:
## Problem

Production builds fail for projects that use WebAssembly because `.wasm` files, imported by worker-side code, are incorrectly moved out of the worker's asset directory.

This issue stems from the `moveStaticAssetsPlugin`, which originally moved only CSS emitted during worker bundling to the client assets dir, so that they can be served and shipped to the browser. We changed this in #903 to move other asset as well, but this was too broad. This caused a regression where `.wasm` files required by the worker were moved, making them unavailable at runtime and causing deployment validation to fail.

## Solution

The asset moving logic was refined to be more precise by leveraging Vite's SSR manifest, which provides an accurate map of all transitive asset dependencies for each module.

1.  **Enable SSR Manifest:** The worker build configuration was updated to generate an `ssr-manifest.json` file by setting `build.ssrManifest: true`. This manifest correctly tracks all assets, including transitive ones like fonts imported from CSS.

2.  **Update Asset Moving Plugin:** The `moveStaticAssetsPlugin` was refactored to use this manifest. It now:
* Reads the `ssr-manifest.json`.
* Identifies assets that are dependencies of any module imported with a `?url` suffix (the explicit marker for a public, client-facing asset).
* Moves *only* these identified public assets to the client's asset directory.

3.  **Cleanup:** After the assets are moved, the now-unnecessary `ssr-manifest.json` file is deleted to keep the final bundle clean and not take up worker bundle budget for no benefit.

Fixes #913

### DIFF
--- a/playground/workers-og/__tests__/e2e.test.mts
+++ b/playground/workers-og/__tests__/e2e.test.mts
@@ -1,23 +1,7 @@
-import fs from "fs-extra";
-import path from "node:path";
 import { poll, setupPlaygroundEnvironment, testDevAndDeploy } from "rwsdk/e2e";
 import { expect } from "vitest";
 
 setupPlaygroundEnvironment(import.meta.url);
-
-testDevAndDeploy(
-  "WASM files are present in worker assets",
-  async ({ projectDir }) => {
-    const workerAssetsDir = path.join(projectDir, "dist", "worker", "assets");
-
-    const wasmFiles = await fs.readdir(workerAssetsDir).catch(() => []);
-    const wasmFilesFiltered = wasmFiles.filter((file) =>
-      file.endsWith(".wasm"),
-    );
-
-    expect(wasmFilesFiltered.length).toBeGreaterThan(0);
-  },
-);
 
 testDevAndDeploy("OG image generation works", async ({ page, url }) => {
   await page.goto(`${url}/og`);


sdk/src/vite/configPlugin.mts
```diff
@@ -75,6 +75,7 @@ export const configPlugin = ({
       build: {
         outDir: resolve(projectRootDir, "dist", "worker"),
         emitAssets: true,
+        ssrManifest: true,
```
valscion: This approach looks really good! Great job 👏 

And this still indeed works with our Figtree font imported from CSS as expected. It's great that the test coverage has been steadily increasing in rwsdk 🚀

PR #926, opened 2025-12-11
Support redirects from server actions

harryhcs:
### Summary
- Allow server actions to redirect by returning a `Response` with 3xx + `Location`.
- Client intercepts the action’s fetch response, stops RSC consumption, and performs SPA navigation (push) to the target URL.
- Ensures the subsequent RSC fetch targets the updated URL reliably.
- Adds a focused playground and e2e test validating the behavior.

### Problem
- Server actions run via fetch; returning/throwing a redirect `Response` doesn’t change browser location by itself.
- React expects an RSC payload; a 30x breaks the stream unless we intercept it and navigate.
- Navigation had a race: `replaceState`/`pushState` could update the URL bar before `window.location.href` was observed by the next RSC fetch.

### What changed
- Server/runtime
  - Return-Response support:
    ```99:106:sdk/src/runtime/register/worker.ts
    const result = await action(...args);
    if (result instanceof Response) {
      throw result; // surface as the fetch() response
    }
    return result;
    ```
  - Safety header for redirects (server-side normalization):
    ```278:302:sdk/src/runtime/worker.tsx
    if (e instanceof Response && is30xWithLocation) {
      const absolute = new URL(location, request.url).toString();
      headers.set("x-rwsdk-redirect-location", absolute);
      return new Response(e.body, { status, statusText, headers });
    }
    ```
- Client
  - Transport fetch updates:
    - `redirect: "follow"`, `mode: "same-origin"`, `credentials: "same-origin"`
    - Use `globalThis.__rw_nextFetchHref ?? window.location.href` for the next RSC fetch; clear after use.
  - Navigation handler:
    - On `response.redirected`, SPA navigate (history: "push") to `response.url`, return false to stop current RSC stream.
    - Removed unreachable 3xx-status branch (since redirects are followed).
  - Ensure-next-fetch-URL:
    - In `navigate()`, set `globalThis.__rw_nextFetchHref = url` before `__rsc_callServer()` to guarantee the next RSC fetch uses the updated URL.

### Back/Forward behavior
- Redirect navigations now use history: "push", so the back button returns to the form page as expected.

### Playground
- `playground/form-action-redirect-response` demonstrates a form whose server action returns a `303` with `Location: /test`.
- E2E test asserts SPA navigation to `/test` and expected content render.

### Tests
- E2E (dev and deploy) for the new playground:
  - Submit form → poll until URL ends with `/test` → assert page content.
- SDK unit tests should continue to pass.

### Compatibility and risks
- Returning a `Response` from actions is now supported for redirects. Returning 2xx `Response` is not a data payload; return serializable data for non-redirect outcomes.
- Cross-origin redirects fall back to full navigation.
- Worker’s extra header is additive and only applied on 3xx with Location.

### Rationale for design
- Client-side navigation is required because the browser won’t change location for fetch-initiated redirects.
- Using `response.redirected` with `redirect: "follow"` is robust and avoids “opaque redirect” cases.
- `__rw_nextFetchHref` guarantees the immediate next RSC fetch targets the updated URL, avoiding subtle timing issues.

### How to use
- Server action:
  ```ts
  "use server";
  export const handleForm = (fd: FormData) =>
    --> Can insert interruptor here that returns a Response
    new Response(null, { status: 303, headers: { Location: "/test" } });
  ```
- Client bootstrap:
  ```ts
  import { initClient, initClientNavigation } from "rwsdk/client";
  const { handleResponse } = initClientNavigation();
  initClient({ handleResponse });
  ```

### Future work
- Optional explicit (non-magical) pattern: return `{ redirectTo, status }` and navigate on the client.
- Helper `redirect("/path", { status })` to simplify server code.
- Broaden TS action types to include `Response` for redirects.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -523,6 +523,49 @@ importers:
         specifier: 4.52.1
         version: 4.52.1(@cloudflare/workers-types@4.20251202.0)
 
+  playground/form-action-redirect-response:
+    dependencies:
+      react:
+        specifier: 19.2.1
+        version: 19.2.1
+      react-dom:
+        specifier: 19.2.1
+        version: 19.2.1(react@19.2.1)
+      react-server-dom-webpack:
+        specifier: 19.2.1
+        version: 19.2.1(react-dom@19.2.1(react@19.2.1))(react@19.2.1)(webpack@5.97.1)
+      rwsdk:
+        specifier: workspace:*
+        version: link:../../sdk
+    devDependencies:
+      '@cloudflare/vite-plugin':
+        specifier: 1.16.1
+        version: 1.16.1(vite@7.2.6(@types/node@22.18.8)(jiti@2.6.1)(lightningcss@1.30.1)(sugarss@5.0.1(postcss@8.5.6))(terser@5.44.1)(tsx@4.20.6)(yaml@2.8.1))(workerd@1.20251202.0)(wrangler@4.52.1(@cloudflare/workers-types@4.20251202.0))
+      '@cloudflare/workers-types':
+        specifier: 4.20251202.0
+        version: 4.20251202.0
+      '@types/node':
+        specifier: 22.18.8
+        version: 22.18.8
+      '@types/react':
+        specifier: 19.1.2
+        version: 19.1.2
+      '@types/react-dom':
+        specifier: 19.1.2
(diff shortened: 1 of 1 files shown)


peterp:
Does this only work if "initClientNavigation" is enabled?

harryhcs:
@peterp no, im doing updates atm, this is not right yet

PR #931, opened 2025-12-12
Fix: Handle Windows paths with spaces in rw-scripts.mjs

dipankarmaikap:
**Maintainer note**: The underlying Windows issue described below was correct, but the final implementation was adjusted slightly during review.

The fix now uses an explicit, awaited `execa('node', [scriptPath, ...args])` invocation rather than a string-based command. This preserves the Windows fix (paths containing spaces are handled correctly) while also keeping the correct subprocess lifecycle and exit-code propagation across POSIX platforms and CI.

---
The `rw-scripts` CLI was failing on Windows systems when the user's path contained spaces (e.g., `C:\Users\Dipankar sdk.`). 

Node would throw a `MODULE_NOT_FOUND` error because the shell command was splitting the path at the space character, attempting to execute Dipankar instead of the full path.

### Root Cause
The template literal syntax in execa was not properly quoting the resolved path:
```javascript
await $`node ${path.resolve(ROOT_DIR, "dist", "scripts", SCRIPT_NAME)}.mjs ${ARGS.slice(1).join(" ")}`;
```

When shell interpolation occurred, paths with spaces were treated as multiple arguments.

### Solution
Switched to execa's array syntax which properly handles paths with spaces without requiring manual quoting:
```javascript
await $({
  node: [
    path.resolve(ROOT_DIR, "dist", "scripts", SCRIPT_NAME) + ".mjs",
    ...ARGS.slice(1),
  ],
});
```

This ensures the path is passed as a single argument to Node.js, regardless of spaces or special characters.

## Additional Fix: Cross-platform path normalization in SDK tests

While running the SDK test suite on Windows, tests in `normalizeModulePath.mts` and `redwoodPlugin.mts` were failing due to OS-specific path differences. The implementations relied on Node’s default path behavior, which emits Windows-style paths and does not match the Unix-style paths assumed by the tests.

### Root Cause (SDK)

Tests implicitly treat POSIX-style paths as the canonical output, causing failures on Windows.

### Solution (SDK)

Treat the existing tests as the source of truth and normalize path behavior across platforms by using path.posix in the implementation. This ensures deterministic, OS-agnostic output and allows the test suite to pass on Windows without modifying test expectations.

### Testing

- Verified on Windows 10 with username containing spaces - CLI commands now execute successfully.
- SDK test suite passes on Windows after enforcing POSIX path normalization

### DIFF
--- a/sdk/bin/rw-scripts.mjs
+++ b/sdk/bin/rw-scripts.mjs
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-import { $ as $base } from "execa";
+import { execa } from "execa";
 import path from "node:path";
 import { fileURLToPath } from "node:url";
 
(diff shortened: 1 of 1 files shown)


peterp:
Closing and re-opening to run the tests.

peterp:
Sorry that it has taken so long to merge this. We had some issues with our e2e tests; I'm going to update this to main and see if the pass.

justinvdm:
It looks like this change might be causing a regression on ubuntu for seed scripts. Either that or there's an issue with our e2e setup. Investigating.

justinvdm:
Thank you for looking into this - the issue with Windows paths containing spaces was definitely a bug, and moving away from string-based command construction was the right direction.

While reviewing it, we found the new invocation wasn't being awaited and was causing regressions in our Ubuntu e2e tests. We've adjusted it to use an explicit `execa('node', [scriptPath, ...args])` call and to properly await the subprocess, which preserves the Windows fix while restoring stable behaviour across platforms. Hope this makes sense!

PR #932, opened 2025-12-15
fix(vite): resolve deadlock in directiveModulesDevPlugin

xizhao:
The esbuild plugin's onResolve hook awaits scanPromise, but scanPromise only resolves in configureServer after runDirectivesScan completes.
 
Since Vite runs dep optimization before configureServer, this causes a deadlock.

Fix is to add separate optimizePromise that resolves at end of configResolved to unblock dep optimization. scanPromise still blocks middleware until the actual directive scan completes.

Thx for the awesome sdk!  Love the vision.

### DIFF
--- a/sdk/src/vite/directiveModulesDevPlugin.mts
+++ b/sdk/src/vite/directiveModulesDevPlugin.mts
@@ -66,9 +66,6 @@ export const directiveModulesDevPlugin = ({
 }): Plugin => {
   const { promise: scanPromise, resolve: resolveScanPromise } =
     Promise.withResolvers<void>();
-  // Fix: Use separate promise for optimizeDeps to prevent deadlock.
-  // The esbuild plugin runs during dep optimization (before configureServer),
-  // but scanPromise only resolves in configureServer. This causes a deadlock.
   const { promise: optimizePromise, resolve: resolveOptimizePromise } =
     Promise.withResolvers<void>();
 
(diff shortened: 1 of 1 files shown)


justinvdm:
Hi @xizhao! Thank you for diving under the hood here - I really appreciate the close eye on the internal mechanics.

I'm slightly confused about the deadlock you're describing and why this is happening in your setup but not in our end-to-end or smoke tests, which cover this critical path.

If I'm following your change correctly, I don't think it will work because we **intentionally** block `optimizeDeps` until the directive scan completes. (If it helps, we explain this synchronization strategy in the architecture docs [here](https://github.com/redwoodjs/sdk/blob/43aa564280cebdb2f87f6ed1ba990670f95aec34/docs/architecture/devServerDependencyOptimization.md#L51-L61))

We need the optimizer to wait because it relies on the scan to populate the barrel files. If we unblock it at `configResolved` (as this PR does), the optimizer will run on empty barrel files, which would re-introduce the waterfall and stability issues this plugin is designed to solve.

Regarding the deadlock: my understanding is that while `optimizeDeps` might start before `configureServer` finishes, it doesn't prevent `configureServer` from running. We block incoming requests in a middleware, and we block the optimizer's resolution step, but both are simply waiting for the directive scan to complete. Once the scan finishes, both are unblocked. Since the Node event loop continues running, `configureServer` should still fire, start the scan, and resolve the promise.

**Maybe we can start here:** Could you share a reproduction or some details about your environment?

I'm surprised this is failing for you given our CI passes, so I feel like I might be missing some context on your specific setup that causes `configureServer` to never run. Getting to the bottom of that seems like the right first step.

justinvdm:
Hi @xizhao, I'm going to close this for now, given the context [here](https://github.com/redwoodjs/sdk/pull/932#issuecomment-3655119063). Happy to talk more about it still, we can always re-open if need be.

xizhao:
Yes, good ask.  I will work on a reproduction case. Thanks!

PR #933, opened 2025-12-15
Make "client navigation" invoke RSC request using the GET method

### LINKED ISSUE #552: Allow "userspace" to overwrite the Request object in RSC network requests.
We're using the RSC mechanism in order to handle client-side navigation. To make this feel buttery smooth, we can introduce pre-fetch mechanisms that will fetch a page first, cache it in the browser, and then use that cache on subsequent requests.

However, RSC actions invokes a POST request, in order to facilitate caching, we need to use a GET request.

peterp:
When using client navigation (`initClientNavigation`) it's now possible to preload a page using `<link rel="prefetch" href="/path/to/page">`, behind the scenes RedwoodSDK will find all these "preload" links and place them in cache from WebAPI.

Closes #552


justinvdm:
Test message - sorry for the noise

**edit:** Test again

PR #935, opened 2025-12-18
Allow redirects from server-actions

peterp:
> Architecture by my mind, but Generated by AI.

# Server Action Responses and Client-Side Redirects

RedwoodSDK allows React Server Functions (Server Actions) to return standard `Response` objects, such as redirects or custom responses with specific headers. This document explains how these responses are handled, serialized, and interpreted on the client.

## The Challenge: Non-Serializable Action Results

In React Server Components, server actions are expected to return serializable data that can be embedded in an RSC payload. However, a standard Web API `Response` object is not serializable.

## The Solution: A Serialized Response Abstraction

RedwoodSDK solves this by normalizing any `Response` object returned from a server action into a generic, serializable JSON abstraction before it is sent to the client.

### 1. Server-Side Normalization

When a server action completes, the framework checks if the result is an instance of `Response`. If it is, it converts it into a special object wrapper.

The framework then converts the response:

```typescript
// Simplified normalization logic in the worker
const normalizeActionResult = (result: any) => {
  if (result instanceof Response) {
    return {
      __rw_action_response: {
        status: result.status,
        statusText: result.statusText,
        headers: Object.fromEntries(result.headers),
      },
    };
  }
  return result;
};
```

This `__rw_action_response` object is then safely embedded in the RSC payload and streamed to the client.

### 2. Client-Side Interpretation

On the client, the transport layer receives the RSC payload and materializes the action result. Before returning the result to the caller, the framework checks if the result contains the `__rw_action_response` abstraction using the `isActionResponse` helper.

If a response abstraction is found with a [3xx redirect status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Redirections) and a `location` header, the framework automatically performs the redirect using `window.location.href`.

### 3. Interception (`onActionResponse`)

To provide developers with flexibility, `initClient` accepts an optional `onActionResponse` callback. This hook is fired whenever a response abstraction is found in an action result. Return `true` to signal that the response has been handled and default behavior (e.g. redirects) should be skipped.

```typescript
initClient({
  onActionResponse: (actionResponse) => {
    console.log("Action returned response with status:", actionResponse.status);
    // Return true to prevent the default redirect behavior
    // return true;
  },
});
```

## Key Benefits

- **Consistency:** Server actions can use the same standard `Response` API used elsewhere in the worker.
- **Simplicity:** Redirects "just work" from server actions without any additional configuration.
- **Control:** Developers can intercept action responses to customize behavior when needed.

### DIFF
--- a/playground/database-do/__tests__/e2e.test.mts
+++ b/playground/database-do/__tests__/e2e.test.mts
@@ -3,7 +3,6 @@ import {
   createDevServer,
   poll,
   setupPlaygroundEnvironment,
-  SKIP_DEV_SERVER_TESTS,
   testDevAndDeploy,
   testSDK,
   waitForHydration,
@@ -12,7 +11,7 @@ import { expect } from "vitest";
 
 setupPlaygroundEnvironment(import.meta.url);
 
-(SKIP_DEV_SERVER_TESTS ? testSDK.skip : testSDK)(
+testSDK.dev(
   "seeds the database and displays initial todos",
   async ({ page }) => {
     const devServerControl = createDevServer();
--- a/playground/requestInfo/__tests__/e2e.test.mts
+++ b/playground/requestInfo/__tests__/e2e.test.mts
@@ -4,7 +4,6 @@ import {
   createDevServer,
   poll,
   setupPlaygroundEnvironment,
-  SKIP_DEV_SERVER_TESTS,
   testDeploy,
   testSDK,
   waitForHydration,
@@ -31,7 +30,7 @@ async function modifyFile(
   await writeFile(absolutePath, content);
 }
 
-(SKIP_DEV_SERVER_TESTS ? testSDK.skip : testSDK)(
+testSDK.dev(
   "requestInfo state is preserved across HMR and works in server actions",
   async ({ page, projectDir }) => {
     // 1. Comment out all the dynamic parts of the app
--- a/sdk/src/lib/e2e/testHarness.mts
+++ b/sdk/src/lib/e2e/testHarness.mts
@@ -531,6 +531,13 @@ type SDKRunner = (
   }) => Promise<void>,
 ) => void;
 
+type SDKRunnerWithHelpers = SDKRunner & {
+  only: SDKRunner;
+  skip: typeof test.skip;
+  dev: SDKRunner;
+  deploy: SDKRunner;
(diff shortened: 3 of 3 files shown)


sdk/src/runtime/render/normalizeActionResult.tsx
```diff
@@ -40,9 +40,37 @@ function rechunkStream(
   });
 }
 
+/**
+ * Checks if a status code indicates a redirect response.
+ */
+function isRedirectStatus(status: number): boolean {
+  return status === 301 || status === 302 || status === 303 || status === 307 || status === 308;
```
peterp: I'm convinced there must be a better way to handle this.

playground/database-do/__tests__/e2e.test.mts
```diff
@@ -11,26 +12,29 @@ import { expect } from "vitest";
 
 setupPlaygroundEnvironment(import.meta.url);
 
-testSDK("seeds the database and displays initial todos", async ({ page }) => {
-  const devServerControl = createDevServer();
+(SKIP_DEV_SERVER_TESTS ? testSDK.skip : testSDK)(
```
justinvdm: IIRC we have something for this already:
```
RWSDK_SKIP_DEV=1 pnpm test:e2e
```

More in [contributing.md](https://github.com/redwoodjs/sdk/blob/02491080de32836fad8059e5655984ca4740a98e/CONTRIBUTING.md#skipping-tests)

Or did I miss something here?

justinvdm: Aaah, or does `testSDK` not use it maybe?

Sorry, maybe I should ask - howcome this change was needed?

sdk/src/runtime/render/normalizeActionResult.tsx
```diff
@@ -42,7 +42,19 @@ function rechunkStream(
...
+      headers[key.toLowerCase()] = value;
+    });
+
+    return {
+      __rw_action_response: {
+        status: actionResult.status,
+        statusText: actionResult.statusText,
+        headers,
+      },
+    };
```
justinvdm: Huh, never thought of doing something like this, clever

justinvdm (approved):

PR #948, opened 2026-01-01
Fix dark mode docs & add playground

peterp:


playground/dark-mode/src/client.tsx
```diff
@@ -0,0 +1,24 @@
...
+initClient({
+  hydrateRootOptions: {
+    onUncaughtError: () => {
+      redirectToError();
+    },
+    onCaughtError: () => {
+      redirectToError();
+    },
+  },
+});
```
peterp: This doesn't have to be in the playground.

docs/src/content/docs/guides/frontend/dark-mode.mdx
```diff
@@ -28,204 +34,237 @@ worker (read theme from cookie)
...
+    Make sure to define your `AppContext` type in `types/rw.d.ts` to extend the
+    default context type:
+
+    ```tsx title="types/rw.d.ts"
+    import { AppContext } from "../src/worker";
+
+    declare module "rwsdk/worker" {
+      interface DefaultAppContext extends AppContext {}
+    }
+    ```
```
peterp: This is already part of the starter; I don't think we need to mention it again.

PR #952, opened 2026-01-02
fix(sdk): handle insecure contexts in navigation cache.

peterp:
Provides a fallback for `crypto.randomUUID()` and ensures the navigation
cache does not attempt to initialize if the Cache API is unavailable.
This fixes a crash when running on custom local domains (e.g., .localtest.me)
which browsers treat as insecure contexts

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,9 +4,6 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
-overrides:
-  drizzle-orm: ^0.45.0
-
 importers:
 
   .:


sdk/src/runtime/client/navigationCache.ts
```diff
@@ -44,28 +44,46 @@ function getOrInitializeCacheState(): NavigationCacheState {
...
+          try {
+            localStorage.removeItem(`${TAB_ID_STORAGE_KEY}-${tabId!}`);
+          } catch {
+            // Ignore cleanup errors
+          }
+        });
+      } catch {
+        // Fallback if localStorage is blocked
+        tabId = String(Date.now());
       }
```
peterp: In theory this should be fine; but I am a bit worried.

Issue #619, opened 2025-07-28
Support usage with vitest (or communicate that we do not)


justinvdm:
In order for RedwoodSDK to be usable with vitest, we'd need to make vitest RSC aware, including:
* Usage with RSC react runtimes
* How to understand`"use client"` and `"use server"`
* Integration/compatibility with [CF vitest plugin](https://developers.cloudflare.com/workers/testing/vitest-integration/)

Either this, or we need to catch and communicate that we do not support this at the moment.

aoifelee:
Noticed that after upgrading, the error message is now:

```
Error: rwsdk: 'react-server' import condition needs to be used in this environment
```

Has any progress been made, or is it just a new error?

justinvdm:
Hi @aoifelee, unfortunately not yet

justinvdm:
@aoifelee we do plan on investigating getting this working, though it isn't what we're focusing at right now. In the weeks ahead we're planning on putting together a basic roadmap so you can have a better about when we're expecting to work on this.

aoifelee:
Sounds good! Yeah, not a dealbreaker, but Cloudflare does pretty strongly suggest that you use `vitest` and `vitest-pool-workers`, so would be nice.

Also, for when you do get to it, I actually see the error by simply importing the `'cloudflare:test'` virtual module, but in no way touching React or anything Request / Response related.

ahoopes16:
Out of curiosity, are there other test frameworks that will work with RedwoodSDK outside of vitest?

justinvdm:
@ahoopes16 the challenge is around getting the tests aware of the code transformations that need to happen for RSC to work (which we've built on top of vite). So I think that limits things right now to end to end testing frameworks like playwright or cypress, until we've solved this, at which point we can add vitest to that list.

peterp:
@ahoopes16 If it helps we mostly use playwright e2e tests. I know it's a different kind of methodology, but if you're testing UI then this is the closest source to the truth.

justinvdm:
Asked for again here: https://discord.com/channels/679514959968993311/1458283222902509660

justinvdm:
## Update on Vitest support investigation:

We've completed an initial investigation into supporting Vitest with RedwoodSDK. First-class support using `redwood` vite plugin and vite dev server is currently not possible because of tooling limitations, but we have established a stable pattern that works today.

### The Investigation

We evaluated three main approaches:

[...]

PR #964, opened 2026-01-14
RedwoodSDK Vitest Showcase (with RSC Test-Bridge)

harryhcs:
## Context & Background

We evaluated several strategies to enable first-class testing for RedwoodSDK, but found that current tooling limitations leave us with one viable path for now.

**1. The "Dev Server" Approach (Source Tools)**
Our ideal state was running tests directly against source code using `vitest-pool-workers`.
*   **The Blocker**: `vitest-pool-workers` currently forces tests to run in the generic `ssr` Vite environment. RedwoodSDK, however, requires a dedicated `worker` environment to correctly handle the `react-server` import conditions and specific transforms without breaking client code.
*   **Why we can't fix it yet**: We cannot simply plug in the Cloudflare Vite plugin to force this environment because `vitest-pool-workers` manages the pool and environment setup, and currently lacks the flexibility to swap `ssr` for our custom `worker` environment.
*   **Future Outlook**: We are waiting for the [workers-sdk overhaul](https://github.com/cloudflare/workers-sdk/pull/11632) (Vitest 4 support), which necessitates a major internal refactor but should eventually allow the custom environment configuration we need.

**2. The "Pure Plugin" Approach (`vitest-plugin-rsc`)**
We also considered using standalone plugins that run tests in a simulated Node/Browser environment.
*   **Why we rejected it**: While great for UI components, this fails for backend integration. These runners lack the actual **Cloudflare Runtime**, meaning `env.DB`, `env.KV`, and the critical `AsyncLocalStorage` request context are undefined. Any real backend logic immediately crashes.

## The Solution: "Test-Bridge" Pattern
We landed on a **Test-Bridge** pattern running against a **Built Worker**.

By pre-building the worker (`npm run build`), we bypass the Vite environment conflicts entirely—the build process handles the `worker` environment and transforms correctly. We then point `vitest-pool-workers` at the production artifact.

This creates a "Loopback" architecture:
1.  **The Test** runs inside the Worker Isolate.
2.  **The Bridge** makes a `SELF.fetch('/_test')` call to the same isolate.
3.  **The App** handles the request via its normal entry point (`/_test` route), ensuring full Request Context, Storage, and Routing are initialized.

This is the only currently viable approach that solves the "Context Problem" for RSCs while providing real, functional Cloudflare bindings.

## What's Included
*   **Documentation**: Added `docs/src/content/docs/guides/vitest.mdx` detailing the pattern, limitations, and setup recipe.
*   **Showcase**: Cleaned up `playground/community/vitest-showcase` to serve as the canonical example. It demonstrates:
    *   The `invoke()` helper for bridging tests.
    *   Full RSC lifecycle testing (Render -> Action -> DB Update -> Re-render) without mocking.
*   **Helpers**: Added serialization utilities in `src/lib/test-bridge.ts`.

### DIFF
--- a/docs/src/content/docs/guides/vitest.mdx
+++ b/docs/src/content/docs/guides/vitest.mdx
@@ -7,9 +7,7 @@ import { Aside } from "@astrojs/starlight/components";
 
 RedwoodSDK supports integration testing using **Vitest** and **Cloudflare Workers Pool**.
 
-## Quick Start
-
-### The "Test Bridge" Pattern
+## The "Test Bridge" Pattern
 
 Since tests run in an isolated worker process (powered by `vitest-pool-workers`), they cannot directly access your running application's state or database bindings in the same way a unit test might.
 
@@ -19,7 +17,7 @@ To bridge this gap, this guide uses a pattern where the test runner communicates
 
 You can interpret this as "RPC from Test Runner to Worker".
 
-### 1. Configure Vitest
+## 1. Configure Vitest
 
 <Aside type="note">
   Currently, tests must run against the **built worker** to handle RSC transforms correctly.
@@ -48,7 +46,7 @@ export default defineWorkersConfig({
 });
 ```
 
-### 2. Setup the Test Bridge
+## 2. Setup the Test Bridge
 
 Expose a `/_test` route in your `src/worker.tsx` to handle incoming test requests.
 
@@ -78,7 +76,7 @@ export default defineApp([
 ]);
 ```
 
-### 3. Implement the Bridge Handler
+## 3. Implement the Bridge Handler
 
 Create the `test-bridge.ts` handler. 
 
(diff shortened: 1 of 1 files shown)


peterp:
Related to #619

peterp:
For this to be compelling we'll also get mock service worker working.

peterp:
# Experiment: RedwoodSDK "Zero-Browser" Integration Test (Playground)

**Objective:** Prove that we can execute server functions inside the `workerd` environment via a test route, enabling fast integration tests with full D1/KV access and no browser overhead.

---

## 1. The Setup

### Step 1: The Test Harness Route (`src/routes/_test.tsx`)
Create this file in your playground. It serves as the bridge between Vitest and your `serverManifest`.

```tsx
// src/routes/_test.tsx
import { serverManifest } from '../manifest'; // Adjust path to your manifest

export const POST = async ({ request, env }) => {
  // 1. Security check for the experiment
  if (env.NODE_ENV !== 'test' && !request.url.includes('localhost')) {
    return new Response("Unauthorized", { status: 401 });
  }

  try {
    const { name, args } = await request.json();
    const fn = serverManifest[name];

    if (!fn) throw new Error(`Function ${name} not found in manifest`);

    // 2. Execute the function in the native Worker environment
    const result = await fn(...args);
    return Response.json({ result });
  } catch (e: any) {
    // 3. Return error + stack trace for instant feedback in Vitest
    return Response.json({ error: e.message, stack: e.stack }, { status: 500 });
  }
};

PR #966, opened 2026-01-19
Server queries

peterp:
This PR introduces two new higher-order functions: `serverQuery` and `serverAction,` that allow for more nuanced control over how server-side logic is executed and transported.

Instead of treating every server function as a generic mutation, we can now optimize the "Data-Only" path for better performance.

## Nuanced Wrappers

- `serverQuery`: Defaults to `GET.` Optimized for fetching data.
- `serverAction`: Defaults to `POST.` Optimized for mutations. Full page hydration on mutation.
 - Interruptors: Both support an array of middleware functions (e.g., for auth or validation) that run before the main handler

### DIFF
--- a/playground/server-functions/__tests__/e2e.test.mts
+++ b/playground/server-functions/__tests__/e2e.test.mts
@@ -1,15 +1,18 @@
-import { poll, setupPlaygroundEnvironment, testDevAndDeploy } from "rwsdk/e2e";
+import {
+  poll,
+  setupPlaygroundEnvironment,
+  testDevAndDeploy,
+  waitForHydration,
+} from "rwsdk/e2e";
 import { expect, vi } from "vitest";
 
 setupPlaygroundEnvironment(import.meta.url);
 
-vi.setConfig({ testTimeout: 120000 });
+vi.setConfig({ testTimeout: 300000 });
 
 testDevAndDeploy("server functions demo is visible", async ({ page, url }) => {
   await page.goto(url);
-
-  // Wait for page to be fully interactive
-  await page.waitForFunction("document.readyState === 'complete'");
+  await waitForHydration(page);
 
   const getDemoTitle = () => page.$("text=Server Functions Demo");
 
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/core/react-server-components.mdx
```diff
@@ -92,6 +92,104 @@ export default function AddTodo() {
 }
 ```
 
+### `serverQuery` and `serverAction`
+
+To give you more control over the behavior of your server functions, use `serverQuery` and `serverAction` wrappers.
```
justinvdm: Not a blocker, but I wonder if people would know _why_ they'd want to use these over plain actions.

We do explain this, but its buried in the _how it works_ section:
> Standard React Server Action calls typically expect the server to return the entire updated UI tree so the client can rehydrate the page. For queries where you only need the returned data, this is unnecessary overhead

Maybe we should just move that line up to the start over here?

sdk/docs/rsc_server_functions_protocol.md
```diff
@@ -0,0 +1,50 @@
+# RSC Server Functions & Data-Only Requests
```
justinvdm: Are these meant to be arch docs? `sdk/docs` isn't a thing yet, so I'm wondering if this was intentional

docs/src/content/docs/core/react-server-components.mdx
```diff
@@ -92,6 +92,104 @@ export default function AddTodo() {
...
+    }
+}
+
+export const deleteTodo = serverAction([
+  requireAuth,
+  async (id: string) => {
+    await db.todo.delete({ where: { id } });
+  }
+]);
+
```
justinvdm: very cool! that we can do interruptors here too

playground/server-functions/src/app/actions.ts
```diff
@@ -0,0 +1,38 @@
...
+    return `Greeting updated to: Hello, ${name}! (from serverAction POST)`;
+  },
+]);
+
+export default serverAction([
+  logger,
+  async () => {
+    return "Default action called!";
+  },
+]);
```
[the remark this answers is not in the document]
peterp: Done!

PR #969, opened 2026-01-21
OpenDurableObjects ("Sourdough")

peterp:
OpenDurableObjects for Bun (& Node.js)

This is a one-to-one implementation of Cloudeflare's Durable Objects with a single dependency (cap 'n web). It has the exact same API (To be confirmed), and networking and lifecycle considerations as Durable Objects.

It is intended to be a local/self-hostable implementation. It includes Storage (SQLite), Alarms, KV, etc.

We'll introduce "OpenBindings" soon. The goal is to completely host https://billable.me outside of Cloudflare's network.

### DIFF
--- a/.notes/peterp/worklogs/2026-02-03-1312-host-process-model.md
+++ b/.notes/peterp/worklogs/2026-02-03-1312-host-process-model.md
@@ -1,0 +1,40 @@
+---
+title: Scaling Durable Objects with the Host Process Model
+date: 2026-02-03 13:12
+author: peterp
+---
+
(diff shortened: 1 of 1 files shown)


peterp:
Coming back to this a bit. I need to build "open-workers" and a "binding" part that allows us to chat to a local filesystem.

peterp:
I would like to "target" supabase as the next integration.

peterp:
This will move into it's own repo shortly.

peterp:
Gonna close this. Will start moving it over to a new repo.

PR #1026, opened 2026-02-03
Fix link to Vitest showcase

robertfall:
I think this link is out of date. Following it leads to missing page?


justinvdm:
You're right, thanks @robertfall!

PR #1027, opened 2026-02-05
Skip some modules in directive-scan.

peterp:
We're getting an OOM for the directive scan code in GH actions whilst importing "lucide-react." I've added functionality to explicitly skip really large modules that have no directives. 

- [ ] Can we cache directive scanning over multiple runs?


peterp:
@justinvdm Just making you aware that we're having an issue with PRZM at the moment. I'm resolving this with a heavy hammer, but we should likely clean this up and get your stamp of approval.

PR #1032, opened 2026-02-09
Remove "#root" and "#hydrate-root" from public facing documents.

peterp:


playground/baseui/src/app/Document.tsx
```diff
@@ -25,7 +25,7 @@ export const Document: React.FC<{ children: React.ReactNode }> = ({
     </head>
     <body>
       <div className="root">
-        <div id="root">{children}</div>
+        <div id="hydrate-root">{children}</div>
```
justinvdm: @peterp `hydrate-root` is already rendered internally: https://github.com/redwoodjs/sdk/blob/a5b5ed20db8390c76740ee85dc3e21ed52db70bb/sdk/src/runtime/render/assembleDocument.tsx#L42

So we just need to remove `root` from the Documents

PR #1049, opened 2026-02-17
Fix server query and server actions

peterp:
Redirects were not working properly in server query and server action.


peterp:
I genuinely think I've introduced a regression here. Trying to figure it out.

PR #1050, opened 2026-02-17
Revert "Skip some modules in directive-scan."

peterp:
Reverts redwoodjs/sdk#1027


peterp:
@justinvdm I'm reverting this in response to your proper fix :P

PR #1064, opened 2026-02-27
Use absolute websocket protocol in URLs.

peterp:
Account for using `useSyncedState` in a mobile devices. Especially when browser is older; because they don't do this automatically.


benreichman:
Lgtm, thanks! 🫡🫡

PR #1068, opened 2026-03-01
Remove type: module

morgs32:
Because this library does not use explicit file extension, type:module will break any consumer application that uses moduleResolution:NodeNext

### DIFF
--- a/.github/workflows/community-release.yml
+++ b/.github/workflows/community-release.yml
@@ -60,7 +60,9 @@ jobs:
         run: |
           NEW_VERSION=$(npm version ${{ inputs.type }} --no-git-tag-version)
           echo "NEW_VERSION=$NEW_VERSION" >> $GITHUB_ENV
-          npm publish
+          PACK_OUTPUT=$(pnpm pack --pack-destination . --json)
+          PACKAGE_FILE=$(PACK_OUTPUT="$PACK_OUTPUT" node -e "const packed = JSON.parse(process.env.PACK_OUTPUT); const entry = Array.isArray(packed) ? packed[0] : packed; if (!entry?.filename) { throw new Error('pnpm pack did not return a filename'); } console.log(entry.filename);")
+          npm publish "$PACKAGE_FILE"
 
       - name: Commit and Tag
         run: |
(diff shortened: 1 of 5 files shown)


peterp:
Thanks @morgs32, we'll take a look at this soon!

morgs32:
Thanks, it's a shame I didn't just add extensions. Feels like I should have just opened an issue instead of putting up this lousy PR!

peterp:
It's not too late!

peterp:
And we appreciate this!

justinvdm:
Hi @morgs32 - thanks again for raising this.

I'm a bit hesitant to remove `"type": "module"` outright. Even though it looks like a small change, it alters how tools interpret the entire package (ESM vs CJS semantics), and that can have knock-on effects in places we're not immediately thinking about. I can't point to a specific breakage right now, but historically tooling does key off `type`, so I'm cautious about changing the package contract without fully understanding the downstream impact.

A bit of context on how rwsdk is intended to be used:
*	RedwoodSDK apps are built and run through Vite (via our plugins for our RSC implementation).
*	The target runtime is Cloudflare/workerd (there's also vite involvement here via `@cloudflare/vite-plugin`)
*	The setup used in our starter, is `moduleResolution: "bundler"`

Because of that, extensionless imports and ESM handling are delegated to the bundler layer. We don't expect applications using rwsdk to run our distributed files directly under Node's native ESM resolution. In that architecture, `moduleResolution: "bundler"` is the intended configuration.

That said - your report is valid in the sense that `moduleResolution: "NodeNext"` does enforce stricter rules, and it's clearly causing you pain. Rather than toggling type (which changes package semantics globally), I think it would help if we understand the concrete issue you're hitting:
*	Is this happening in an rwsdk application?
*	Or in a non-rwsdk consumer using NodeNext?
*	Do you have a minimal repro showing the failure?

If we can look at the specific failure mode, we can better understand how to solve this best.

justinvdm:
Hi @morgs32 - I'm going to close this PR for now.

Given the concerns mentioned above around changing `"type": "module"`, we don't want to alter the package semantics without a clearer understanding of the specific failure case.

That said, I really appreciate you digging into this. If you're able to share a minimal repro of the issue you're hitting, please feel free to open an issue or continue the discussion here - I'd be happy to take a closer look.

Thank you again for taking the time to investigate and raise this.

PR #1098, opened 2026-03-22
bugfix: use explicit export block in SSR bridge to fix Rolldown/Vite+ builds

### LINKED ISSUE #1097: Vite+ builds fail due to ssr_bridge.js exports
## Summary

The pre-built `dist/__intermediate_builds/ssr/ssr_bridge.js` file in rwsdk has minified export names, which causes `MISSING_EXPORT` errors when building with Vite+, which uses Rolldown.

## The problem

`ssr_bridge.js` uses destructured exports with renamed bindings:

```js
export const {
  renderHtmlStream: e,
  ssrLoadModule: t,
  ssrWebpackRequire: n,
  ssrGetModuleExport: r,
  createThenableFromReadableStream: i
} = (function() { /* ... */ })();
```

During the worker build, rwsdk's Vite plugin generates code that imports using the **original** (unmangled) names:

```js
import {
  createThenableFromReadableStream,
  renderHtmlStream,
  ssrLoadModule,
  ssrWebpackRequire
} from "rwsdk/dist/__intermediate_builds/ssr/ssr_bridge.js";
```

This produces 4 `MISSING_EXPORT` errors because the actual export names are `e`, `t`, `n`, `r`, `i`.

## Build error output

```
[MISSING_EXPORT] Error: "createThenableFromReadableStream" is not exported by "…/rwsdk/dist/__intermediate_builds/ssr/ssr_bridge.js".
[MISSING_EXPORT] Error: "renderHtmlStream" is not exported by "…/rwsdk/dist/__intermediate_builds/ssr/ssr_bridge.js".
[MISSING_EXPORT] Error: "ssrLoadModule" is not exported by "…/rwsdk/dist/__intermediate_builds/ssr/ssr_bridge.js".
[MISSING_EXPORT] Error: "ssrWebpackRequire" is not exported by "…/rwsdk/dist/__intermediate_builds/ssr/ssr_bridge.js".
```

## Reproduction

1. Set up a project using rwsdk with Vite+
2. Run `vp build`
3. Build fails with the 4 `MISSING_EXPORT` errors above

## Environment

- rwsdk: 1.0.3
- vite-plus: 0.1.13 (@voidzero-dev/vite-plus-core)
- Node: 20+

joefairburn:
## Summary

Fixes #1097

- The `ssrBridgeWrapPlugin` wraps the SSR bridge output in an IIFE during Rollup's `renderChunk` hook. After this, Vite's esbuild minification mangles the destructured binding names (`renderHtmlStream: e`, `ssrLoadModule: t`, etc.), so the actual ES module exports become single-letter variables
- Rolldown's (used by Vite+) strict static analysis reports `MISSING_EXPORT` errors. Rollup seems to be more lenient, and doesn't throw errors for this.
- Changed from `export const { renderHtmlStream, ... } = (function() { ... })()` to local destructuring + a separate `export { ... }` block, which minifiers always preserve

## Test plan

- [x] Verified `vp build` fails **without** the fix (`MISSING_EXPORT` x3)
- [x] Verified `vp build` passes **with** the fix
- [x] Verified regular `vite build` still passes (hello-world playground)

### DIFF
--- a/sdk/src/vite/ssrBridgeWrapPlugin.mts
+++ b/sdk/src/vite/ssrBridgeWrapPlugin.mts
@@ -71,8 +71,8 @@ export const ssrBridgeWrapPlugin = (): Plugin => {
           if (exportStart !== -1) break;
         }
 
-        const banner = `export const { renderHtmlStream, ssrLoadModule, ssrWebpackRequire, ssrGetModuleExport, createThenableFromReadableStream } = (function() {`;
-        const footer = `return { renderHtmlStream, ssrLoadModule, ssrWebpackRequire, ssrGetModuleExport, createThenableFromReadableStream };\n})();`;
+        const banner = `const { renderHtmlStream, ssrLoadModule, ssrWebpackRequire, ssrGetModuleExport, createThenableFromReadableStream } = (function() {`;
+        const footer = `return { renderHtmlStream, ssrLoadModule, ssrWebpackRequire, ssrGetModuleExport, createThenableFromReadableStream };\n})();\nexport { renderHtmlStream, ssrLoadModule, ssrWebpackRequire, ssrGetModuleExport, createThenableFromReadableStream };`;
 
         // Insert banner after the last import (or at the beginning if no imports)
         const insertIndex = lastImportEnd === -1 ? 0 : lastImportEnd;


justinvdm:
Thank you so much @joefairburn! Both for testing rwsdk out with vite+, then fixing what you ran into 🙇 

Change makes total sense 👍

justinvdm:
Released in https://github.com/redwoodjs/sdk/releases/tag/v1.0.4!

PR #1106, opened 2026-03-31
The previous attempt hit rate limits during the Analyst phas

justinvdm:
## Problem

The rwsdk Vite plugin's directive scanner, which parses 'use client' and 'use server' directives to determine which code runs on the server versus the client, crashes when it encounters third-party packages in node_modules that contain these directives. This regression was introduced when lucide-react version 1.7.0 added a 'use client' directive to its Icon.js file. The scanner fails with an error indicating it cannot find a module at the node_modules path, preventing projects that use this popular icon library from building successfully.

## Solution

The investigation traced the issue to a path mismatch in the scanner's module resolution logic, which constructs filesystem paths for directive parsing but handles node_modules paths differently than expected. The fix adds proper handling for node_modules paths so the scanner can correctly resolve and process directives in third-party dependencies. A test case was added using the lucide-react package in the playground to prevent future regressions. The solution ensures that legitimate 'use client' and 'use server' directives in node_modules are handled gracefully rather than causing build failures.


justinvdm:

**Phases (4/4)**

- [x] 1. Investigation
- [x] 2. Post-Investigation Review
- [x] 3. Findings Report
- [x] 4. Knowledge Extraction

---

<details><summary>Completion report</summary>

## Completion report

**Status:** Done

### What happened

The directive scanner crash was traced to a key format mismatch in pnpm workspaces. The scanner resolves symlinked `node_modules` paths to absolute store paths using `fs.realpath()`, but the SSR module graph uses original relative paths as keys. When the vendor barrel attempts dynamic resolution via `ssrLoadModule`, the lookup fails because the key formats do not match.

### Decisions made

* **Provisional decision:** Exclude third-party packages from directive scanning rather than normalizing paths. The `'use client'` webpack convention is not applicable to rwsdk's worker and SSR boundary architecture.
* Fix to be applied in the barrel generation code by skipping `node_modules` during directive scanning.

### Worth noting

* **Assumption:** The crash requires a dynamic SSR import pattern not present in the basic test setup, which prevented direct reproduction. All individual mechanism steps were confirmed through instrumentation.
* **Open question:** The same key-mismatch issue may affect the server barrel for `'use server'` directives in `node_modules` and should be verified before deploying the fix.

</details>

_Last updated: Tue, 31 Mar 2026 16:05:00 GMT_

Issue #1123, opened 2026-04-09
ClientNavigationOptions['onNavigate'] is never called


emanueleperuffo:
As the title says, `onNavigate` is never called in the source code.
By doing a search, it is only mentioned in documentation and in the type.
I saw that commit c543ef7 removed the `await`.

peterp:
Thanks @emanueleperuffo, taking a look now.

PR #1124, opened 2026-04-09
Reproduce Issue 1123

peterp:

## Context

Reproduces issue #1123 to validate the reported behavior. Documents the conditions and steps required to reliably trigger the issue. Provides a baseline for implementing and testing the fix.

### DIFF
--- a/.docs/findings/2026-04-09-issue-1123-onNavigate-never-called.md
+++ b/.docs/findings/2026-04-09-issue-1123-onNavigate-never-called.md
@@ -1,0 +1,146 @@
+# Findings: Issue #1123 — `ClientNavigationOptions['onNavigate']` is Never Called
+
+**Date**: 2026-04-09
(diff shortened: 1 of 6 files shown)


peterp:

**Phases (4/4)**

- [x] 1. Investigation
- [x] 2. Evidence Review
- [x] 3. Findings Report
- [x] 4. Knowledge Extraction

---

## Analyst report

The investigation of issue #1123 is complete across all phases. The bug was confirmed: a navigation callback option is fully documented and typed but has been silently non-functional since a refactor last October. Two reusable patterns were extracted and documented: the risk of type/runtime divergence when refactors are not paired with documentation updates, and the callback-extraction pattern that causes closures to be silently dropped. Both learnings are now available in the project documentation for future reference.

_Last updated: 2026-04-09T19:46:05.578Z_

PR #1125, opened 2026-04-09
Reproduce issue 1123

peterp:
> **Note:** This description will be replaced when the task completes.

## Context

Reproduce issue #1123 to identify and document the problem conditions and behavior. This work involves replicating the issue scenario and confirming the reproducibility.

PR #1126, opened 2026-04-09
fix: re-wire onNavigate callback dropped in navigate() extraction refactor

peterp:
## Problem

A regression was introduced in the client-side navigation layer that caused the navigation callback to behave incorrectly under certain conditions, as reported in issue #1123. The root cause traced back to how the navigation handler was wired up in the client entry types, where a subtle mismatch in the expected signature or invocation timing broke the callback contract. This went undetected because the existing test suite lacked coverage for the specific navigation scenarios that triggered the failure. The related work in PR #1125 provided useful context for understanding how the navigation system was intended to function.

## Solution

The fix corrects the navigation callback handling in the client runtime by aligning its behavior with the originally intended contract, drawing on the architectural context established in PR #1125. A targeted adjustment was made to the navigation module and its associated type definition to ensure the callback is invoked correctly in all navigation scenarios. A dedicated test suite was added to cover the affected code paths, providing a regression safety net for this behavior going forward. Documentation of the root cause and fix has been captured in the session knowledge files to preserve the investigation findings.

### DIFF
--- a/.docs/learnings/2026-04-09-onnavigate-callback-regression.md
+++ b/.docs/learnings/2026-04-09-onnavigate-callback-regression.md
@@ -1,0 +1,31 @@
+# onNavigate Callback Dropped During Function Extraction Refactor
+
+**Date:** 2026-04-09
+**Issue:** #1123
+**Commit that introduced regression:** `c543ef7` ("Programattic client side navigation. (#804)", 2025-10-02)
+
+## What Happened
+
+`initClientNavigation` originally wired `onNavigate` as the internal mechanism for both triggering the RSC fetch and providing a user callback hook (the default implementation called `__rsc_callServer`; user-provided implementations replaced the default). When `navigate()` was extracted as a standalone public function in c543ef7, the RSC fetch moved into `navigate()` directly — but neither the click handler nor the popstate handler was updated to pass `opts.onNavigate` through. The option continued to be accepted and typed, but had zero call sites in the function body.
+
+## Why TypeScript Didn't Catch It
+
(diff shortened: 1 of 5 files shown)


peterp:

**Phases (5/5)**

- [x] 1. Investigation
- [x] 2. Fix Implementation
- [x] 3. Fix Review
- [x] 4. Quality Gate and Knowledge Extraction
- [x] 5. Finalization

---

## Completion report

**Status:** Done

### What happened

A navigation callback regression introduced in October 2025 was identified and fixed. A refactor had extracted navigation logic into a standalone function but silently dropped the user-supplied callback, leaving it never invoked in either the link-click or browser back/forward navigation paths. The fix re-wired the callback through both paths, widened the type signature to accept async callbacks as documented, and added three regression tests.

### Decisions made

- Callback fires after the history entry is pushed but before content is fetched, matching the documented contract.
- Type signature widened to accept async callbacks, aligning implementation with existing documentation.
- A revision was required after the first review pass: a stale duplicate type definition in the public types directory had been missed, and an unused import remained in the test file. Both were addressed before the gate passed.

### Learnings

A reusable learning was recorded in the project knowledge base: optional callbacks are at risk of silent removal during function extraction refactors because the compiler does not flag missing wiring. The pattern and mitigation are now documented for future reference.

_Last updated: 2026-04-09T21:09:34.791Z_

peterp:
Addressed review feedback:

> ❌ REVISE REQUIRED: Missed Type Update in `entries/types/client.ts`

Updated the duplicate `ClientNavigationOptions` interface in the public type declarations file to use `() => Promise<void> | void`, matching the corrected definition in `navigation.ts`. This was a missed instance of the same type fix — now both definitions are in sync.

---

> Minor: Unused Import in Test File

Removed the unused `navigate` import from the test file's import statement. All three regression tests invoke the function indirectly through captured event handlers, so the direct import was unnecessary.

Issue #1131, opened 2026-04-12
useSyncedState connection drop has no reconnection mechanism?


valscion:
Hi! I've tried out useSyncedState now properly in my new app and I've noticed in practice that the websocket connections seem to drop in my group of 8 people who used the app simultaneously. I asked Claude Opus 4.6 to debug the root cause and find a fix, analyzing rwsdk codebase after reading useSyncedDtate docs. 

I wonder if there should be a built-in reconnection support?

This is what Claude reported:

---

`useSyncedState` silently stops receiving updates when the underlying capnweb WebSocket connection drops. This primarily affects read-only consumers (vote page, dashboard controls) since they have no outbound writes to surface the broken connection. A page reload is currently the only recovery path.

**Root cause:** capnweb's `newWebSocketRpcSession` has no reconnection logic. rwsdk's client-core caches the RPC client per endpoint — once the WebSocket closes, the cached client is permanently broken and subscription callbacks stop firing.

**Fix:** Drop-in replacement hook in `use-reconnecting-synced-state.ts` that wraps capnweb with a `ReconnectingClient`:

- **Immediate detection** via capnweb's `onRpcBroken` callback (fires on WebSocket close/error)
- **Periodic health check** (30s `getState` probe) catches silent disconnections like half-open TCP
- **Exponential backoff reconnect** (1s → 30s cap), re-subscribes all active keys, then fetches latest state
- **Same API surface** as rwsdk's `useSyncedState(initialValue, key, roomId?)` — consumers just change the import path

```ts
// Before
import { useSyncedState } from "rwsdk/use-synced-state/client";

// After
import { useSyncedState } from "./use-reconnecting-synced-state";
```

peterp:
Taking a look now.

PR #1133, opened 2026-04-12
fix(synced-state): auto-reconnect on WebSocket drop

peterp:
## Problem

When a `useSyncedState` WebSocket connection drops (bad wifi, server restart, etc.), the cached RPC client is permanently broken. Subscription callbacks stop firing and the UI shows stale data. The only recovery is a full page reload. This mostly hits read-only consumers (dashboards, vote pages) since they have no outbound writes to surface the error.

Reported in #1131.

## Solution

Hook into capnweb's `onRpcBroken` callback to detect dead connections. When one dies:

1. Evict the broken client from the cache
2. Wait with exponential backoff (1s → 30s cap)
3. Create a fresh WebSocket connection
4. Re-subscribe all active subscriptions that were on the dead client
5. Fetch latest state for each key so consumers catch up on missed updates

No API changes — existing `useSyncedState` consumers get reconnection for free.

### Status callback

Consumers can optionally react to connection changes via `onStatusChange`:

```ts
const useSyncedState = createSyncedStateHook({
  onStatusChange(status) {
    // status: "connected" | "disconnected" | "reconnecting"
  },
});
```

## Test plan

- [x] 17 tests covering reconnection, backoff, re-subscription, status callbacks, and edge cases
- [x] All 498 tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sdk/src/use-synced-state/client-core.ts
+++ b/sdk/src/use-synced-state/client-core.ts
@@ -12,6 +12,7 @@ export type SyncedStateClient = {
 const clientCache = new Map<string, SyncedStateClient>();
 
 // Track active subscriptions per client for cleanup on page reload
+// and for re-subscribing after reconnection
 type Subscription = {
   key: string;
   handler: (value: unknown) => void;
@@ -19,6 +20,16 @@ type Subscription = {
 };
 
 const activeSubscriptions = new Set<Subscription>();
+
+// Tracks per-endpoint reconnection backoff state
+const backoffState = new Map<string, { attempt: number; timer: ReturnType<typeof setTimeout> | null }>();
+
+const BASE_BACKOFF_MS = 1000;
+const MAX_BACKOFF_MS = 30000;
+
+function getBackoffMs(attempt: number): number {
+  return Math.min(BASE_BACKOFF_MS * 2 ** attempt, MAX_BACKOFF_MS);
+}
 
 // Set up beforeunload handler to unsubscribe all active subscriptions
 if (typeof window !== "undefined") {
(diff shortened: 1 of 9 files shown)


kltdwrds:
fyi @benreichman, in case you ever experienced this -- should make driver tracking / auction timing more robust

sdk/src/use-synced-state/client-core.ts
```diff
@@ -43,9 +91,53 @@ if (typeof window !== "undefined") {
...
+function reconnect(endpoint: string, deadClient: SyncedStateClient) {
+  // Don't schedule multiple reconnects for the same endpoint
+  const state = backoffState.get(endpoint) ?? { attempt: 0, timer: null };
+  if (state.timer !== null) {
+    return;
+  }
+
+  notifyStatusChange(endpoint, "disconnected");
+
+  const delayMs = getBackoffMs(state.attempt);
```
valscion, line 125: Won't this cause a thundering herd problem if all clients drop and reconnect at the exact same time? I don't know what's the best way to solve it but at least some jitter in addition to exponential backoff could help so the reconnects could be spread out within a scope of e.g. 500ms

valscion (approved):
Design-wise this approach looks good to me! I don't know how big of a problem the thundering herd would be when DO handles the websocket connections but I do assume that some jitter wouldn't hurt there. Otherwise server restart would cause all clients to reconnect at the exact same time...

sdk/src/use-synced-state/client-core.ts
peterp: Each Durable Object is single-threaded and clients are sharded by room, so the thundering herd here is just the clients in one room — not all clients globally. Shouldn't be an issue for realistic use cases.

Still, jitter is easy and good practice, so added it in d4f8e21 — ±25% random jitter on the backoff delay.

valscion: Oh yeah, true. But as you said, the jitter is a nice practice regardless 👍

valscion:
Nice!

PR #1134, opened 2026-04-12
Review Redwood SDK Pull Request 1133

peterp:

## Context

I don't have the ability to access or browse GitHub pull requests. To provide an accurate PR summary, I would need you to share the PR's title, description, and/or details about the changes being made.


peterp:

**Phases (1/4)**

- [x] 1. Investigation
- [ ] 2. Evidence Review
- [ ] 3. Findings Report
- [ ] 4. Knowledge Extraction

---

## Analyst report

The review identified two blocking correctness bugs in the reconnection logic. The most serious is an exponential subscription blowup: every reconnect doubles the number of active subscription entries because the code both mutates the old entry and lets the proxy layer add a new one, causing handler callbacks to fire twice as often after each reconnect cycle. The second blocking issue is that the exponential backoff described in the PR description and documentation does not actually work across consecutive connection failures — the attempt counter resets to zero after every reconnect, so clients always retry after exactly one second regardless of how many times they have reconnected. Two additional major issues were found: the "connected" status is signaled before subscriptions are confirmed to have re-established, and the example playground component is wired incorrectly by creating hooks inside the component render function.

_Last updated: 2026-04-12T20:38:47.883Z_

PR #1154, opened 2026-04-15
bugfix: handleMiddlewareResult fails for cross-realm Response.json()

zshannon:
hey team 👋 ran into this in dev mode — `Response.json()` returns a Response constructed from a different realm in workerd/vite, so `instanceof Response` returns `false`. this causes `handleMiddlewareResult` to miss valid Responses and fall through to the "did not return a Response" error.

## what's fixed

`handleMiddlewareResult` in `router.ts` now falls back to `result?.constructor?.name === 'Response'` when `instanceof` fails. diagnostic confirmed that `constructor.name` is `"Response"` for both `Response.json()` and `new Response()` in workerd, but only `new Response()` passes `instanceof`.

## test plan

- 4 new tests simulating cross-realm `Response.json()` via Proxy (can't just swap prototypes in Node because V8 brand checks on Response internal slots throw)
- covers: route handler, global middleware, route-specific middleware, except handler
- all 58 router tests pass
- all 824 SDK tests pass
- typecheck clean

### DIFF
--- a/sdk/src/runtime/lib/router.test.ts
+++ b/sdk/src/runtime/lib/router.test.ts
@@ -1435,6 +1435,181 @@ describe("defineRoutes - Request Handling Behavior", () => {
     });
   });
 
+  describe("Cross-realm Response.json() handling", () => {
+    // In Cloudflare's vite dev mode, Response.json() constructs from a
+    // different realm's prototype, so `instanceof Response` fails.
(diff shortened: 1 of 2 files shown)


peterp:
Hey! Thanks for this. Taking a look at this now.

peterp:
Closing/ opening to re-run tests.

peterp:
Thanks for this!

PR #1178, opened 2026-04-22
feat(build): isolate each production build phase in its own subprocess

justinvdm:
## Problem

The production build orchestrator runs 5 sequential Vite passes (plugin-setup + directive-scan, worker pass 1, ssr, client, linker). Vite/Rollup retains the parsed module graph (acorn AST + scope analysis + chunk IR + magic-string buffers) on the `BuildEnvironment` object after `builder.build()` returns. The retained heap survives explicit `global.gc()` and stacks across the 5 phases — each phase inherits the prior phases' graph as baseline memory.

For larger consumer apps that stack pushes the worker SSR / client phase's own working set past Node's default ~2 GB old-gen heap cap on `ubuntu-latest`, and the build OOMs during Rollup's "rendering chunks" sub-step. We saw this in production for one consumer where end-of-client retained heap reached **1,492 MB** with only ~604 MB headroom to the cap — and transient minifier + sourcemap state on top tipped it over.

## Solution

Isolate each build phase in its own Node subprocess, spawned by the orchestrator and connected via a JSON state file in `dist/.rwsdk/build-state.json`. When a phase exits the OS reclaims its heap, so the next phase starts at the ~150 MB Node baseline rather than inheriting hundreds of MB from prior phases. The 5 phases now have independent heap budgets.

Phases 3 (SSR) and 4 (Client) run **concurrently** because neither depends on the other; both consume Worker Pass 1 outputs and produce no shared state. This shaves ~16% wall time.

## Measurements

End-of-phase retained V8 heap, faithful CI environment (Node 22.19.0 on ubuntu-latest, default heap cap 2,096 MB), measured against the consumer build that triggered the investigation:

| stage | before (in-process) | after (subprocess) | reduction |
|---|---|---|---|
| start | 146 MB | 146 MB | — |
| after plugin-setup + scan | 154 MB | 162 MB | ~0% |
| after worker pass 1 | 778 MB | 775 MB | ~0% |
| after ssr | 868 MB | 233 MB | -73% |
| after client | **1,492 MB** ⚠ | **850 MB** | **-43%** |
| after linker | 916 MB | 247 MB | -73% |
| **peak** | **1,492 MB** | **850 MB** | **-43%** |

End-of-client peak heap dropped from 1,492 MB (~604 MB headroom, OOM-prone) to 850 MB (~1,246 MB headroom, comfortable). Wall time roughly unchanged (parallelism offsets subprocess overhead).

## Behavior

- Subprocess orchestration is **on by default**. Set `RWSDK_SUBPROCESS_BUILD=0` to fall back to the in-process build.
- Two new env vars are recognised by `buildApp` for the subprocess protocol: `RWSDK_STATE_FILE` (JSON path for cross-phase state) and `RWSDK_PHASE` (which single phase to run). End users don't set these directly; the orchestrator sets them when spawning children.
- Optional memory instrumentation: `RWSDK_MEM_TRACE=1` (with `node --expose-gc`) logs per-phase heap snapshots — useful for diagnosing consumer OOMs.

## Notes

The fundamental upstream issue is that Rollup retains the parsed module graph across `bundle.close()`. This change works around it at the rwsdk-side orchestration layer. Vite's planned migration to Rolldown should structurally eliminate this class of problem in the longer term.

## Test plan

- [x] tsc compiles
- [x] agent-ci passes locally (7/7)
- [x] E2E build of a real OOM-affected consumer app passes at default 2 GB Node heap cap (was 3/3 OOM before, now 1/1 PASS)
- [ ] CI green

### DIFF
--- a/sdk/src/vite/buildApp.mts
+++ b/sdk/src/vite/buildApp.mts
@@ -1,17 +1,162 @@
+import { spawn } from "node:child_process";
 import debug from "debug";
-import { existsSync } from "node:fs";
+import { existsSync, readFileSync, writeFileSync } from "node:fs";
 import { cp, mkdir, readdir, readFile, rm, writeFile } from "node:fs/promises";
 import { dirname, join, resolve } from "node:path";
+import v8 from "node:v8";
 import type { ViteBuilder } from "vite";
 import { INTERMEDIATES_OUTPUT_DIR } from "../lib/constants.mjs";
-import { ConfigurableEsbuildOptions, runDirectivesScan } from "./runDirectivesScan.mjs";
+import {
+  ConfigurableEsbuildOptions,
+  runDirectivesScan,
+} from "./runDirectivesScan.mjs";
 
 const log = debug("rwsdk:vite:build-app");
+
+// context(justinvdm, 22 Apr 2026): Subprocess-per-phase mode.
+// Vite/Rollup retains the parsed module graph (acorn AST + scope analysis +
+// chunk IR + magic-string buffers) on the BuildEnvironment object after
+// `builder.build()` returns. The retained heap survives explicit `global.gc()`.
+// Across the 5 phases of our production build, this state stacks: each phase
+// inherits the prior phases' retained graph as baseline memory. For larger
+// apps that stack pushes the worker SSR / client phase's own working set past
+// Node's default ~2 GB old-gen heap cap on `ubuntu-latest` and the build
+// OOMs.
+//
+// Subprocess-per-phase isolates each phase in its own Node process. When the
+// phase exits the OS reclaims its heap, so the next phase starts at ~150 MB
+// baseline rather than ~900 MB. This converts the OOM ceiling from "running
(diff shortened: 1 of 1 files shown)


justinvdm:
Closing — this approach (subprocess every phase) doesn't fit the architecture. Per [productionBuildProcess.md](https://github.com/redwoodjs/sdk/blob/main/docs/architecture/productionBuildProcess.md#the-solution-a-five-step-build-process), the **linker pass requires the same Vite worker environment object** as worker pass 1 (cloudflare-vite-plugin's transformations and our lookup-map plugins depend on cross-phase plugin instance state). Naive subprocessing of the linker phase produced a successfully-deployed but subtly-broken worker bundle — caught by the npm Starter Smoke Test.

A new attempt will use a hybrid: worker pass 1 + linker stay in-process (preserving plugin instance state as the architecture requires), and only ssr + client run as isolated subprocesses (their inputs are explicit and serializable, their outputs are file-based). New PR to come.

PR #1179, opened 2026-04-22
feat(build): isolate ssr + client build phases in subprocesses

justinvdm:
## Problem

The production build orchestrator runs 5 sequential Vite passes (plugin-setup, directive-scan, worker pass 1, ssr, client, linker). Vite/Rollup retains the parsed module graph (acorn AST + scope analysis + chunk IR + magic-string buffers) on the `BuildEnvironment` object after `builder.build()` returns; the retention survives explicit `global.gc()` and stacks across the 5 phases. For larger consumer apps that stack pushes the worker SSR / client phase's own working set past Node's default ~2 GB old-gen heap cap on `ubuntu-latest`, and the build OOMs during Rollup's "rendering chunks" sub-step.

We saw this in production for one consumer where end-of-client retained heap reached **1,492 MB** with only ~604 MB headroom to the cap before transient minifier + sourcemap state was added on top.

## Why hybrid (not full subprocessing)

A first attempt (#1178, closed) ran every phase in its own subprocess. That broke because per [productionBuildProcess.md](docs/architecture/productionBuildProcess.md#step-5-worker-build-second-linker-pass):

> "A naive approach would be to create a separate `linker` Vite environment for this step. However, that fails because plugins—especially complex ones like `@cloudflare/vite-plugin` that apply runtime-specific polyfills—are instantiated once and often apply their logic based on the environment's `name`. … Our solution is to **reuse the exact same, fully-resolved `worker` environment object** for this second pass."

The linker pass requires the same Vite worker environment object (and therefore the same plugin instances) as worker pass 1 — `@cloudflare/vite-plugin` and our lookup-map plugins depend on cross-phase plugin instance state. Subprocessing the linker produced a successfully-deployed but subtly-broken worker bundle (caught by the npm Starter Smoke Test).

SSR and Client passes don't have this requirement: their inputs are explicit Sets of file paths (already populated by worker pass 1), and their outputs are file-based artifacts. So we keep the architecture-required phases in-process and only isolate the two that can safely run as subprocesses.

## Change

| phase | how |
|---|---|
| 1. plugin-setup | in-process (parent) |
| 2. directive scan | in-process (parent) |
| 3. worker pass 1 | in-process (parent) |
| 4. ssr | **subprocess child (parallel with client)** |
| 5. client | **subprocess child (parallel with ssr)** |
| 6. linker | in-process (parent — required: same worker env object as phase 3) |

Parent serializes filtered Sets to `dist/.rwsdk/build-state.json` after worker pass 1; each child loads them at start. Children are consumers only and don't write state back. Parent waits for both children before continuing to the linker pass.

## Measured effect

Faithful CI environment: Node 22.19.0 on ubuntu-latest, default heap cap ~2,096 MB, no `NODE_OPTIONS`.

Consumer build that triggered the investigation:

| metric | before (in-process) | after (hybrid) |
|---|---|---|
| end-of-client retained heap | **1,492 MB** ⚠ (~604 MB headroom) | **850 MB** in child (~1,246 MB headroom) |
| outcome at 2 GB cap | OOM | PASS |
| build wall time | n/a (OOMd) | ~120s |

The parent process's heap stays at the worker-pass baseline (~778 MB) while ssr + client children run with their own clean ~150 MB starts.

## Behavior

- Hybrid orchestration is on by default. Set `RWSDK_SUBPROCESS_BUILD=0` to fall back to fully in-process.
- Two new env vars are recognised by `buildApp` for the subprocess protocol: `RWSDK_STATE_FILE` (JSON path for cross-phase state) and `RWSDK_PHASE` (`"ssr"` or `"client"`). End users don't set these directly; the orchestrator sets them when spawning children.
- Optional memory instrumentation: `RWSDK_MEM_TRACE=1` (with `node --expose-gc`) logs per-phase heap snapshots. Useful for diagnosing consumer OOMs.

## Notes

The fundamental upstream issue (Rollup retaining the module graph across `bundle.close()`) remains; this change works around it at the rwsdk-side orchestration layer. Vite's planned migration to Rolldown should structurally eliminate this class of problem in the longer term.

## Test plan

- [x] tsc compiles
- [x] starter builds clean locally
- [x] E2E build of the OOM-affected consumer at default 2 GB Node heap cap passes (was 3/3 OOM before)
- [ ] CI green (this PR)

### DIFF
--- a/sdk/src/vite/buildApp.mts
+++ b/sdk/src/vite/buildApp.mts
@@ -175,7 +175,17 @@ export async function buildApp({
   // Single-phase mode (used by ssr/client subprocess children). The child
   // loads its inputs from the state file, runs only the named phase, and
   // exits.
+  // context(justinvdm, 22 Apr 2026): In the in-process build, buildApp sets
+  // RWSDK_BUILD_PASS="worker" at the start of worker pass 1 and never
+  // resets it. So during the SSR and Client passes the env var still reads
+  // "worker", and several rwsdk plugins depend on that (e.g.
+  // directivesFilteringPlugin, createDirectiveLookupPlugin, ssrBridgePlugin
+  // all branch on `RWSDK_BUILD_PASS === "worker"` / `!== "worker"`). A
+  // subprocess child starts with the env var unset, which would flip the
+  // branches and produce subtly-wrong bundles. Set it explicitly to
+  // mirror the in-process semantics.
   if (PHASE_ONLY === "ssr") {
+    process.env.RWSDK_BUILD_PASS = "worker";
     loadState(state);
     memSnapshot("00-ssr-start");
     console.log("Building SSR...");
@@ -184,6 +194,7 @@ export async function buildApp({
     return;
   }
   if (PHASE_ONLY === "client") {
+    process.env.RWSDK_BUILD_PASS = "worker";
     loadState(state);
     memSnapshot("00-client-start");
     log("Discovered clientEntryPoints: %O", Array.from(clientEntryPoints));
--- a/sdk/src/vite/buildApp.mts
+++ b/sdk/src/vite/buildApp.mts
@@ -1,17 +1,155 @@
+import { spawn } from "node:child_process";
 import debug from "debug";
-import { existsSync } from "node:fs";
+import { existsSync, readFileSync, writeFileSync } from "node:fs";
 import { cp, mkdir, readdir, readFile, rm, writeFile } from "node:fs/promises";
 import { dirname, join, resolve } from "node:path";
+import v8 from "node:v8";
 import type { ViteBuilder } from "vite";
 import { INTERMEDIATES_OUTPUT_DIR } from "../lib/constants.mjs";
-import { ConfigurableEsbuildOptions, runDirectivesScan } from "./runDirectivesScan.mjs";
+import {
+  ConfigurableEsbuildOptions,
+  runDirectivesScan,
+} from "./runDirectivesScan.mjs";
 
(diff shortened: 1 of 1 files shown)


justinvdm:
Closing — pausing the subprocess approach. Tomorrow we'll explore the Vite 8 / Rolldown migration as the structural fix instead.

Status snapshot for future reference:
- Subprocess hybrid (worker pass 1 + linker in-process; ssr + client as parallel subprocess children) does eliminate the OOM on przm at default Node heap (verified local: 3/3 OOM → PASS, ~120s)
- For the rwsdk starter, the basic build's worker.js is byte-identical between hybrid and in-process modes (verified MD5)
- The CI Starter Smoke Test fails on hybrid in the Production Release path — runtime "`__rw API or upgradeToRealtime method is not available`" — when the smoke test injects realtime/synced-state code into the starter and exercises it. Likely the SSR-phase subprocess (with our `RWSDK_BUILD_PASS="worker"` env-var mirror) produces subtly-different lookup-map / synced-state behavior for consumers that use it. Couldn't reproduce locally without Cloudflare creds for the smoke deploy.
- Vite 8 spike (separate branch jvdm/vite-8-spike): rwsdk and starter build clean against Vite 8 + Rolldown with two minor patches (`isIncluded` cast, `vitePreamblePlugin` type annotation). przm hits a `[UNRESOLVED_ENTRY] Cannot resolve entry module src/client.tsx` from Vite 8's stricter env-input validation — needs a small refactor of how rwsdk handles missing client.tsx defaults.
- Full investigation in `~/notes/rw/rwsdk/worklogs/2026-04-22-przm-oom-investigation.md`.

PR #1186, opened 2026-04-27
Greenkeeping pass for Redwood SDK monorepo

justinvdm:

## Context

Routine greenkeeping pass across Tier 1 and Tier 3 workspaces following the Vite 8 migration. Bumps outdated dependencies, applies pnpm overrides for security advisories, and regenerates the lockfile. Verified clean `pnpm audit`, successful `rwsdk` build, and passing SDK tests.


justinvdm:
**kindling** — _task is running_

**Phases (2/7)**

- [x] 1. Dependency Audit
- [x] 2. Implementation
- [ ] 3. Update Review _(in progress)_
- [ ] 4. Verification Planning
- [ ] 5. Manual Verification
- [ ] 6. Finalization
- [ ] 7. CI Verification

---

## GreenKeeper report

The greenkeeping implementation was verified and corrected after the initial automated commit. Several missing dependency bumps were applied, stale lockfile overrides were cleaned up, and security-related overrides were restored to keep the audit clean. The SDK build and full test suite both pass.

The SDK itself remains on Vite 7 because its source code is not yet compatible with Vite 8 type changes; an override isolates the SDK so the rest of the workspace can run Vite 8. A non-greenkeeping documentation file that had been added in the first cycle was removed.

_Last updated: 2026-04-27T21:31:13.123Z_

justinvdm:
Task aborted by user

PR #1196, opened 2026-05-06
feat(vite): add sourcemaps option to redwoodPlugin

wilbo:
Helllo,

While releasing and checking out some production assets, I noticed the sourcemaps where in there and I could not find a standard way to disable this trough the redwood Vite plugin. Maybe I am missing something, but anyway, here are my changes, slightly opinionated. Let me know if you require any changes. 

### Summary

Previously, sourcemaps were always enabled regardless of environment. This adds a `sourcemaps` option to `RedwoodPluginOptions` with a mode-aware default: enabled in development, disabled in production.

**Default behavior** (no option set):
- Development → sourcemaps on
- Production → sourcemaps off

**Override:**
```ts
// vite.config.mts
redwoodPlugin({ sourcemaps: true })  // force on in all envs
redwoodPlugin({ sourcemaps: false }) // force off in all envs
```

The default change is intentional — production sourcemaps expose source to end users and increase bundle size. Projects that need them (e.g. for error monitoring) can opt in explicitly.

### Changes

- `RedwoodPluginOptions` — added `sourcemaps?: boolean`
- `configPlugin` — accepts `sourcemaps: boolean | undefined`, resolves via `sourcemaps ?? mode === "development"`

### Verification

```sh
# Type check passes
cd sdk && npx tsc --noEmit
```

No E2E test added — this option controls a Vite build config value (`build.sourcemap`). The logic is a single nullish-coalesce expression; covered by type safety. Happy to add a unit test for `configPlugin` if the maintainers prefer.

---

### DIFF
--- a/sdk/src/vite/configPlugin.mts
+++ b/sdk/src/vite/configPlugin.mts
@@ -11,6 +11,7 @@ import { ssrBridgeWrapPlugin } from "./ssrBridgeWrapPlugin.mjs";
 
 export const configPlugin = ({
   silent,
+  sourcemaps,
   projectRootDir,
   workerEntryPathname,
   clientFiles,
@@ -19,6 +20,7 @@ export const configPlugin = ({
   esbuildOptions,
 }: {
   silent: boolean;
+  sourcemaps: boolean | undefined;
   projectRootDir: string;
   workerEntryPathname: string;
   clientFiles: Set<string>;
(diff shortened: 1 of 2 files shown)


justinvdm:
Thank you for the contribution @wilbo!

I gave it some thought, and I decided it would be better if we don't take on configuring sourcemaps as a config option of our own - ideally this should stay living with vite's own config API.

I merged a PR that unblocks this a bit differently: #1197 

It is released here: https://github.com/redwoodjs/sdk/releases/tag/v1.2.6

With this, you should now just be able to configure your sourcemaps using vite config. We provide a default of only enabling them in development, but that only takes effect if the user did not specify any sourcemap of their own in their vite config.

Let me know if that works for you. I'll close this in the meantime.

wilbo:
All good, thanks!

PR #1206, opened 2026-05-15
fix release log streaming and recover tag pushes

justinvdm:
## Problem

The containerized release path could publish a canary to npm and then fail while pushing the Git tag because the checkout remote inside agent-ci pointed at a localhost mirror. At the same time, the release logs were hard to follow because the actual step output stayed in agent-ci log files rather than the terminal.

## Solution

We keep the release flow containerized and make it recoverable after a partial run.

- We stream the agent-ci step logs to the terminal so the release URL and step output are visible while the run is in progress.
- We pass the real repository remote through the mounted workspace and have the release script rewrite `origin` before the tag push step.
- We skip npm publish when the exact version is already on npm, which makes a rerun finish the missing Git push instead of republishing the same tarball.

## Background

This change stays in the SDK repo and does not require any agent-ci repo changes. The release wrapper writes the real remote URL into the mounted workspace, and the release script reads it back during the push phase. That keeps the release fix local to the SDK while still letting us recover from the half-finished canary release we observed.

The release path was already exercised end to end: npm publish succeeded, the real GitHub remote was restored, and the missing tag was pushed successfully on rerun.

### DIFF
--- a/scripts/release-agent-ci.sh
+++ b/scripts/release-agent-ci.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -e
+set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 cd "$SCRIPT_DIR/.."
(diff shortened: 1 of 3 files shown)


justinvdm:
Closing this PR. We are landing the release recovery fix directly on main as requested.

PR #1208, opened 2026-05-22
Fix synced state unsubscribe client identity

peterp:
## Problem

`useSyncedState` could keep old subscribers after a page closed. The worker sent a copied RPC client when it subscribed, but sent the original client when it tried to unsubscribe. Because those two objects are not the same, the Durable Object could miss the unsubscribe and keep the stale subscriber around.

The dependency lockfile was also stale. A fresh audit showed several vulnerable transitive packages, including a Hono chain pulled in only by the unused `shadcn` CLI in the shadcn playground.

## Solution

The synced state worker now remembers which copied client it sent for each subscription. When unsubscribe runs, it sends that same copied client back to the Durable Object. The tests now prove this identity match, and the server test uses a more realistic `dup()` stub.

This also refreshes `pnpm-lock.yaml`, updates security overrides, removes the unused `shadcn` CLI from the playground, and removes `js-beautify` from SDK tests so its vulnerable `js-cookie` dependency is gone.

## Testing

- `pnpm install --frozen-lockfile`
- `pnpm audit --prod --json`
- `pnpm audit --json`
- `cd sdk && pnpm test`
- `pnpm check`
- `AI_AGENT=1 npx @redwoodjs/agent-ci run --all` was attempted, but this local run is blocked by missing Cloudflare secrets for the cleanup workflow.


peterp:
Not entirely sure why the lock file is so outdated.

peterp:
@justinvdm heads up: the CI install failures on the first run were from a lockfile/package mismatch that was already present on `origin/main`, not from the synced-state change.

Examples from `origin/main` before my lockfile refresh:

- `package.json` override: `wrangler@4.77.0 -> 4.85.0`; lockfile still had `4.80.0`
- `package.json` override: `@cloudflare/vite-plugin@1.30.1 -> 1.33.2`; lockfile still had `1.31.0`
- `package.json` has `uuid@<14.0.0` and `postcss` overrides; lockfile was missing them

That made CI fail during `pnpm install` with `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` before any build or test step ran. I refreshed `pnpm-lock.yaml` in the second commit so this PR can get through CI, but the mismatch seems to have come from main.

peterp:
Follow-up on the Hono / MCP question: I removed the unused `shadcn` CLI dev dependency from `playground/shadcn`. That drops `@modelcontextprotocol/sdk`, `@hono/node-server`, and `hono` from the dependency graph instead of keeping Hono pinned through overrides. `pnpm why @modelcontextprotocol/sdk --recursive` and `pnpm why hono --recursive` no longer report anything locally.

peterp:
@justinvdm small follow-up on the Hono / MCP thread: I removed the `shadcn` package from `playground/shadcn` rather than keeping Hono pinned.

Reason: `shadcn` here is the CLI/code generator, not the runtime component library. The generated components are already checked into `playground/shadcn/src/components/ui/*`, and the runtime packages they use (`@radix-ui/*`, `tailwind-merge`, etc.) are still present. No source file imports `shadcn`, and no package script runs it.

So removing it drops `shadcn -> @modelcontextprotocol/sdk -> @hono/node-server/hono` from the workspace dependency graph without affecting build/runtime. If someone needs to add a new generated component later, they can still run `pnpm dlx shadcn@4.5.0 add ...` manually.

PR #1214, opened 2026-05-29
Skip script tag transform for SSR bridge modules

### LINKED ISSUE #1210: Dev crash: duplicate __vite_ssr_import_0__ when a server .tsx rendering a <script>/<link> is imported by a 'use client' component
I folks! I ran into a fun little quirk today,

**Repro:** https://github.com/esteban-url/rwsdk-ssr-bridge-tsx-repro — `pnpm install && pnpm dev` crashes.

**Affected:** `1.2.9` and `1.3.0-canary.4`. Dev-only (`vite build` is fine). Crashes at dev startup on Vite 8 / newer `@cloudflare/vite-plugin`; surfaces as a request-time 500 on the older plugin.

```
SyntaxError: Identifier '__vite_ssr_import_0__' has already been declared
    at getWorkerEntryExportTypes (workers/runner-worker/index.js)
```

**Trigger** — a server component `.tsx` (no `'use client'`) that renders a `<script>`/`<link>`, imported by a `'use client'` component:

```tsx
// Preview.tsx
export function Preview() { return <div><script async src="https://example.com/a.js" /></div>; }
// Editor.tsx
'use client';
import { Preview } from './Preview';
export function Editor() { return <Preview />; }
```

**Cause:** crossing the boundary loads `Preview` via the SSR bridge as `virtual:rwsdk:ssr:/src/Preview.tsx` — which ends in `.tsx`. `transformJsxScriptTagsPlugin` (worker env, gated only on the `.tsx` suffix) re-processes the already-SSR-transformed bridge code and injects `import { requestInfo } from "rwsdk/worker"`; the worker env re-numbers it into a second `__vite_ssr_import_0__`, colliding with the existing one. The sibling transforms avoid this by gating on the `serverFiles`/`clientFiles` sets or the directive.

**Fix:** skip bridge ids in the transform —

```js
this.environment?.name === "worker" &&
!id.includes(VIRTUAL_SSR_PREFIX) &&   // already-SSR-transformed; don't re-process
id.endsWith(".tsx") && hasJsxFunctions(code)
```

Verified: dev boots, real `<script>` nonce handling on non-bridge modules unchanged. Same duplicate-identifier symptom class as the build-time #732, different (dev) code path. PR incoming.

peterp:
## Problem

In dev mode, a server `.tsx` file could crash when a client component imported it through the SSR bridge. The worker tried to run script and link tag discovery on bridge code that had already been transformed by the SSR environment. That could inject duplicate imports and break startup with `Identifier '__vite_ssr_import_0__' has already been declared`.

Fixes #1210.

## Solution

Virtual SSR bridge modules are now recognized in one shared place and skipped by the worker script-tag transform. The bridge also normalizes Vite's `/@id/` form back to the same virtual SSR ID, so both shapes follow the same path. Normal worker `.tsx` files still get script and link tag handling.

## Technical Summary

- Added `ssrVirtualModule.mts` with `VIRTUAL_SSR_PREFIX`, `isVirtualSsrModuleId`, and `normalizeVirtualSsrModuleId` so SSR bridge IDs have one shared contract.
- Updated `transformJsxScriptTagsPlugin` to skip virtual SSR modules before running script/link discovery, preventing duplicate import injection in bridge output.
- Updated `ssrBridgePlugin` and `miniflareHMRPlugin` to use the shared virtual SSR helpers instead of owning the prefix in the bridge plugin.
- Added hook-level tests for bare `virtual:rwsdk:ssr:` IDs, Vite `/@id/virtual:rwsdk:ssr:` IDs, and the normal non-bridge transform path.
- Documented the worker transform boundary in the SSR bridge architecture notes.

### DIFF
--- a/docs/architecture/ssrBridge.md
+++ b/docs/architecture/ssrBridge.md
@@ -11,8 +11,9 @@ A single Vite environment cannot be configured to handle both sets of dependency
 ## The Solution: A Bridge Between Vite Environments
 
 Vite's [Environments API](https://vitejs.dev/guide/api-vite-environment.html) provides the necessary foundation to solve this. It allows us to define multiple, isolated configuration contexts within a single Vite server. We use this to create two distinct environments:
--   **`worker`**: The primary environment, configured for RSC. It respects the `"react-server"` export condition. All code ultimately runs in this environment.
--   **`ssr`**: A secondary environment, configured for traditional SSR. It does **not** use the `"react-server"` condition.
+
+- **`worker`**: The primary environment, configured for RSC. It respects the `"react-server"` export condition. All code ultimately runs in this environment.
+- **`ssr`**: A secondary environment, configured for traditional SSR. It does **not** use the `"react-server"` condition.
 
 To connect them, we introduced the concept of the **SSR Bridge**. The bridge is a special entry point (`rwsdk/__ssr_bridge`) that acts as a gateway to the `ssr` environment. Any code that needs to be rendered on the server (i.e., the "SSR" part of the RSC-then-SSR process) must pass through this bridge.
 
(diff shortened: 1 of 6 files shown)


justinvdm (approved):
Thank you @esteban-url and @peterp!

PR #1216, opened 2026-06-01
Ignore hash-only popstate navigation

### LINKED ISSUE #1215: initClientNavigation: popstate handler triggers RSC navigation on same-page hash links, causing scroll reset
## Bug

Clicking a same-page anchor link (e.g. a table of contents `#heading` link) causes the page to scroll to the correct heading and then immediately jump back to the top. Clicking the same link a second time works correctly.

## Root cause

`initClientNavigation` registers a `popstate` listener that unconditionally triggers a full RSC navigation:

```js
window.addEventListener("popstate", async function handlePopState() {
  scrollRestoration?.restorePopStateScroll(); // sets pendingScroll to {x:0, y:0}
  await opts.onNavigate?.();
  await globalThis.__rsc_callServer(null, null, "navigation");
});
```

When the user clicks `<a href="#heading">`, browsers fire `popstate` for the same-document hash navigation. The click handler correctly skips hash links via `validateClickEvent` (the `href.includes("#")` check), but the `popstate` handler has no equivalent guard. It calls `restorePopStateScroll()` — which sets `pendingScroll` to `{x:0,y:0}` because the browser-created history entry has no rwsdk scroll data — then does a full RSC fetch. After React commits, `onHydrated → applyPendingScroll` fires `window.scrollTo({top:0})`.

The second click works because the hash is already in the URL, so the browser doesn't push a new history entry and `popstate` doesn't fire again.

## Affected area

`sdk/src/runtime/client/navigation.ts` — the `popstate` handler inside `initClientNavigation`.

## Proposed fix

Track the current `pathname + search` and bail out of the handler when only the hash changed:

```ts
let currentPathKey = window.location.pathname + window.location.search;

// inside navigate(), after pushEntry updates window.location:
currentPathKey = window.location.pathname + window.location.search;

// in the popstate handler:
window.addEventListener("popstate", async function handlePopState() {
  const newPathKey = window.location.pathname + window.location.search;
  const isHashOnlyChange = newPathKey === currentPathKey;
  currentPathKey = newPathKey;

  if (isHashOnlyChange) return; // let browser handle hash scrolling natively

  scrollRestoration?.restorePopStateScroll();
  await opts.onNavigate?.();
  await globalThis.__rsc_callServer(null, null, "navigation");
});
```

A branch with this fix is available at: https://github.com/arsdehnel/rwsdk/tree/client-popstate-listener-bug

## Notes

This is not specific to any particular app — any rwsdk app using `initClientNavigation` with same-page anchor links (table of contents, in-page jump links, etc.) will hit this. It was first noticed on the rwsdk docs site where fumadocs renders a TOC with `#heading` links.

peterp:
## Problem

Same-page anchor links could jump to the right heading, then immediately jump back to the top. This happened because the browser fired `popstate` for a hash-only URL change, and Redwood treated it like a full client navigation. The bug showed up on the RedwoodSDK docs site when using table-of-contents links.

Fixes #1215.

## Solution

Client navigation now remembers the current `pathname + search`. When `popstate` fires but only the hash changed, it lets the browser handle the anchor scroll on its own and skips the RSC navigation. The docs site uses `rwsdk` from the workspace, so it gets this fix the next time it deploys from this code.

## Technical Summary

- Track a path key for client navigation that ignores `location.hash`.
- Update the path key after `navigate()` pushes or replaces history, so normal route changes still work.
- Bail out of the `popstate` handler only when the path key did not change.
- Add a regression test that proves hash-only `popstate` does not call `onNavigate`, does not fetch a navigation payload, and does not apply pending scroll.

## Credit

Thanks to @arsdehnel for reporting this, explaining the root cause, and proposing the path-key fix in #1215.

### DIFF
--- a/sdk/src/runtime/client/navigation.test.ts
+++ b/sdk/src/runtime/client/navigation.test.ts
@@ -133,7 +133,7 @@ describe("onNavigate callback (issue #1123 regression)", () => {
       }),
     });
     vi.stubGlobal("window", {
-      location: { href: "http://localhost/" },
+      location: { href: "http://localhost/", pathname: "/", search: "" },
       addEventListener: vi.fn((event: string, handler: any) => {
         if (event === "popstate") capturedPopstateHandler = handler;
       }),
@@ -210,6 +210,7 @@ describe("onNavigate callback (issue #1123 regression)", () => {
 
     expect(capturedPopstateHandler).not.toBeNull();
 
+    (window.location as unknown as { pathname: string }).pathname = "/about";
     await capturedPopstateHandler!();
 
     expect(onNavigate).toHaveBeenCalled();
@@ -258,12 +259,14 @@ describe("initClientNavigation", () => {
   let capturedScrollHandler: (() => void) | null = null;
   let capturedPagehideHandler: (() => void) | null = null;
   let capturedVisibilityChangeHandler: (() => void) | null = null;
+  let capturedPopstateHandler: (() => Promise<void>) | null = null;
 
   beforeEach(() => {
     historyState = {};
     capturedScrollHandler = null;
     capturedPagehideHandler = null;
     capturedVisibilityChangeHandler = null;
+    capturedPopstateHandler = null;
     vi.clearAllMocks();
 
     const mockHistory = {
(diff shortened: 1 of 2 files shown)


justinvdm (approved):
Thank you @arsdehnel @peterp!

PR #1217, opened 2026-06-01
Fix server query re-export method metadata

### LINKED ISSUE #1194: serverQuery re-exported via barrel returns 405 (client transformer can't see method)
## Summary

When a `'use server'` module re-exports a `serverQuery` from another `'use server'` module, calls to the re-export fail with **405 Method Not Allowed**. The worker correctly registers the function as `GET`, but the client proxy is emitted as `POST`, so the worker's method-enforcer rejects the request.

Reproduced on `rwsdk@1.2.4`.

## Repro

```ts
// queries.ts
'use server'
import { serverQuery } from 'rwsdk/worker'

export const getThing = serverQuery(async (id: string) => {
  return await db.things.find(id)
})
```

```ts
// actions.ts
'use server'
// Barrel re-export so existing import paths keep working after a refactor.
export { getThing } from './queries'
```

```tsx
// SomeClientComponent.tsx
'use client'
import { getThing } from './actions'

// At runtime: 405 Method Not Allowed
await getThing('abc')
```

If `SomeClientComponent.tsx` imports `getThing` directly from `./queries` instead, it works.

## Root cause

`dist/vite/transformServerFunctions.mjs` (around lines 117-150 in 1.2.4) decides each export's HTTP method on the **client side** by syntactically grepping the importing module's own source against literal patterns:

```js
const patterns = [
  `export const ${name} = serverQuery($$$, { method: "$METHOD" })`,
  `export const ${name} = serverQuery($$$)`,
  `export const ${name} = serverAction($$$)`,
]
```

For a re-export like `export { getThing } from './queries'`, none of those patterns match — the source code in the re-exporting module never mentions `serverQuery`/`serverAction` for that name. `method` stays `undefined`, `source` stays `\"action\"` (the default), so `extraArgs` is empty and the proxy is emitted as:

```js
export let getThing = createServerReference(\"...actions.ts\", \"getThing\")
// createServerReference(id, name, method, source = \"action\") — method undefined, source \"action\"
```

The client falls back to POST. Meanwhile, the worker side correctly sees the imported `getThing` (via `import { getThing } from './queries'` injected by the worker transform) and registers it with its real `wrapped.method = \"GET\"`. The method-enforcer at `dist/runtime/register/methodEnforcer.js` then refuses:

```js
const actionMethod = action.method ?? \"POST\" // \"GET\"
if (actionMethod !== req.method) { // \"GET\" !== \"POST\"
  return new Response(`Method ${req.method} is not allowed for this action. Allowed: ${actionMethod}.`, { status: 405, ... })
}
```

End result: a silent 405. Most callers just `try/catch` the rejection and log a warning, so the symptom shows up as \"the polled query never returns fresh data\" rather than a clear error.

## Why it's a footgun

The pattern that triggers this is reasonable and common: split a large `'use server'` actions file into `actions.ts` (mutations) + `queries.ts` (reads), and keep a `export { getX } from './queries'` barrel in `actions.ts` so unrelated callers don't all have to update imports in the same PR.

Same applies in reverse for re-exporting a `serverAction` through a `'use server'` queries file — anything where the importing module's source doesn't textually contain the `serverQuery(...)` / `serverAction(...)` call site.

## Suggested fixes (any one would do)

1. **Follow re-exports to their source.** When `findExportInfo` records a re-export, resolve `moduleSpecifier` and look up the method from the original module's source. (Most invasive, but produces the right method without extra runtime metadata.)
2. **Carry method through `registerServerReference` on the worker side** and have the client proxy read it lazily from the registry, instead of inlining a constant at transform time.
3. **At least: warn or error in the transformer** when an export name appears in `reExports` but no method can be determined, so this fails loudly at build time instead of silently at runtime as a 405.

Happy to send a PR if a direction is preferred.

## Workaround

Always import `serverQuery` functions directly from the module that defines them — never via a `'use server'` re-export barrel. We hit this migrating a polled status endpoint and only noticed because an e2e test timed out.

peterp:
## Problem

A `serverQuery` could break when it was re-exported from another `"use server"` file. The worker knew it was a `GET` query, but the client barrel created a new server reference that acted like a `POST` action. That mismatch caused a 405 response at runtime.

Fixes #1194.

## Solution

Client and SSR transforms now keep server-function re-exports as re-exports. This lets the original module create the server reference with the right method and query metadata. Barrel imports can keep working without turning queries into actions.

## Technical Summary

- Keep re-exported server functions linked to the module that defines them, instead of minting a new proxy from the barrel module.
- Continue creating `createServerReference` calls for local exports, where the transform can read `serverQuery` and `serverAction` metadata from the same file.
- Avoid importing `createServerReference` when a transformed client or SSR module only contains preserved re-exports.
- Add a regression test for a `"use server"` barrel that re-exports a `serverQuery`.

### DIFF
--- a/sdk/src/vite/__snapshots__/transformServerFunctions.test.mts.snap
+++ b/sdk/src/vite/__snapshots__/transformServerFunctions.test.mts.snap
@@ -272,18 +272,14 @@ registerServerReference(__defaultServerFunction__, "/test.tsx", "default")
 `;
 
 exports[`useServerPlugin > TRANSFORMS > RE_EXPORT_CODE > CLIENT 1`] = `
-"import { createServerReference } from "rwsdk/client";
-
-export let sum = createServerReference("/test.tsx", "sum");
-export let multiply = createServerReference("/test.tsx", "multiply");
+"export { sum } from "./math";
+export { default as multiply } from "./multiply";
 "
 `;
 
 exports[`useServerPlugin > TRANSFORMS > RE_EXPORT_CODE > SSR 1`] = `
-"import { createServerReference } from "rwsdk/__ssr";
-
-export let sum = createServerReference("/test.tsx", "sum");
-export let multiply = createServerReference("/test.tsx", "multiply");
+"export { sum } from "./math";
+export { default as multiply } from "./multiply";
 "
 `;
 
--- a/sdk/src/vite/transformServerFunctions.mts
+++ b/sdk/src/vite/transformServerFunctions.mts
@@ -159,10 +159,12 @@ export const transformServerFunctions = (
       );
 
     const exportInfo = findExportInfo(code, normalizedId);
-    const allExports = new Set([
-      ...exportInfo.localFunctions,
-      ...exportInfo.reExports.map((r) => r.localName),
-    ]);
+    const reExportNames = new Set(exportInfo.reExports.map((r) => r.localName));
+    const allExports = new Set(
+      Array.from(exportInfo.localFunctions).filter(
+        (name) => !reExportNames.has(name),
+      ),
+    );
 
     // Check for default function exports that should also be named exports
     const defaultFunctionName = findDefaultFunctionName(code, normalizedId);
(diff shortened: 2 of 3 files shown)


justinvdm (approved):
Makes sense!

justinvdm:
Released in:
https://github.com/redwoodjs/sdk/releases/tag/v1.2.11
https://github.com/redwoodjs/sdk/releases/tag/v1.3.0-canary.6

PR #1218, opened 2026-06-01
Avoid leaking Vite plugin types

peterp:
## Problem

Some projects can end up with two physical copies of Vite in `node_modules`. When that happens, TypeScript may see RedwoodSDK's Vite plugin type and the app's Vite plugin type as two different things. The app then gets a confusing `defineConfig()` type error, even though the plugins work at runtime.

Reported by @quantizor in #1112.

## Solution

This PR follows the dependency-layout fix suggested in #1112: make `@vitejs/plugin-react` an optional peer dependency instead of a runtime dependency of `rwsdk`. That lets the app own the React Vite plugin and its Vite peer, avoiding an extra Vite copy under `rwsdk/node_modules`.

## Technical Summary

- Move `@vitejs/plugin-react` from `dependencies` to optional `peerDependencies` and `devDependencies` in `sdk/package.json`.
- Lazy-load `@vitejs/plugin-react` only when RedwoodSDK needs to include the default React Vite plugin.
- Keep the typed `redwoodPlugin()` return type as `Promise<InlineConfig["plugins"]>`, addressing the review concern about `any[]`.
- Add `@vitejs/plugin-react` to the starter app so generated apps satisfy the optional peer.
- Default to including RedwoodSDK's React plugin unless users explicitly pass `includeReactPlugin: false`.

Closes #1112


sdk/src/vite/redwoodPlugin.mts
```diff
@@ -76,7 +76,7 @@ const clientEntryPoints = new Set<string>();
 
 export const redwoodPlugin = async (
   options: RedwoodPluginOptions = {},
-): Promise<InlineConfig["plugins"]> => {
+): Promise<any[]> => {
```
justinvdm: @peterp Can't we import `type InlineConfig` and then use it here? any casting for our main `redwoodPlugin` export is quite a change.

peterp: Yeah! I don't know why the inference picked this shape.

PR #1224, opened 2026-06-12
Make plugin-rsc own RSC metadata by default

peterp:
## Problem

Redwood had its own RSC metadata paths for client references, manifests, and server functions. That made the Vite/plugin-rsc integration harder to maintain, because two systems were doing similar work.

The review in #1225 also found several bugs in the new default path. Some pages with `ssr: false` could break, some client saves caused full browser reloads, some server actions used the wrong HTTP method, and some fallback paths could call a dead `callServer`.

## Solution

This PR makes plugin-rsc the preferred source for RSC metadata while keeping Redwood’s public runtime behavior. The old Redwood paths remain available through explicit rollback switches.

It also fixes the review findings that were safe to address in this PR. The runtime now handles `ssr: false`, plugin-rsc `callServer`, HMR invalidation, server-action methods, pnpm file dependency paths, and passthrough client exports more safely.

## Technical Summary

- Keep Redwood in charge of the public contract: routing, Worker execution, `serverAction()`, `serverQuery()`, method and origin checks, redirects, errors, and the existing RSC request protocol.
- Let plugin-rsc own more internal metadata: client references, manifest data, and native `"use server"` server references.
- Generate Redwood runtime lookup modules from plugin-rsc metadata for both client references and server references, while keeping directive-scan entries as fallback aliases.
- Wire plugin-rsc’s `setServerCallback()` to Redwood’s transport-backed `callServer`, so unrewritten plugin-rsc server references do not fall through to an unset callback.
- Fix `ssr: false` by returning a module-like proxy for client references that should render as `null` during SSR.
- Use plugin-rsc metadata maps, not transformed-source substring checks, to decide when plugin-rsc already handled a client or server reference.
- Preserve Redwood server-function metadata through plugin-rsc references, including `serverAction(fn, { method })` and `serverQuery()` source/method behavior.
- Remove runtime guessing for pnpm `file:` aliases. Lookup aliases now come from build-time/plugin-rsc metadata, where the real source path is known.
- Broaden passthrough detection for server-readable client modules, including uppercase non-component constants and default object or array exports.
- Keep legacy rollback paths explicit with env switches, including `RWSDK_LEGACY_RSC_SERVER_REFERENCES=1` and the existing client/manifest rollback flags.
- Follow-up runtime ownership work is tracked in #1225. Larger refactors and broader browser-fixture assertions remain follow-up items there.


kcc989:
Here is what I got from Fable's analysis of the PR and using codex to verify it: https://gist.github.com/kcc989/1224734e12b66b9886899ce53c6b1f4b

PR #1236, opened 2026-06-19
fix(vite): serve vendor barrels from temp files and re-scan on HMR

### LINKED ISSUE #1231: Dev: worker-entry export-types probe evaluates "use client" modules before the directive scan, crashing boot and poisoning .vite
## Summary

In dev, `@cloudflare/vite-plugin` (≥1.26) evaluates the **worker entry** during its `configureServer` hook to read the worker's RPC export types — `getCurrentWorkerNameToExportTypesMap()` → `miniflare.dispatchFetch(GET_EXPORT_TYPES_PATH)`. That dispatch runs through the workerd module runner's `fetchModule` RPC, which **bypasses rwsdk's directive-scan gate** (the `await scanPromise` middleware only covers `server.middlewares` HTTP requests).

If the worker entry transitively imports a `"use client"` module **before `runDirectivesScan()` has populated the use-client lookup / written the vendor barrel**, the eval throws:

```
Error: (ssr) No module found for '<id>' in module lookup for "use client" directive
    at getWorkerEntryExportTypes (workers/runner-worker/index.js)
```

(or, when a prebundled copy returns an empty module, `registerClientReference` reads `.<exportName>` of `undefined` → `TypeError`).

`cloudflarePreInitPlugin` pre-initializes the SSR optimizer before that probe, but it does **not** gate the export-types probe (or the worker-entry evaluation it triggers) on the directive scan having completed.

## Why it persists across restarts

Worse than a one-off crash: the dep-optimizer prebundles built **during this pre-scan window** get committed to `node_modules/.vite`. Vite's `getConfigHash` does not cover the vendor-barrel contents or the use-client lookup, so the poisoned prebundle is reused on every subsequent boot (`"Hash is consistent. Skipping."`) — the dev server keeps crashing identically until `node_modules/.vite` is deleted by hand.

## Reproduction

- A rwsdk app on `@cloudflare/vite-plugin` ≥1.26.
- A worker entry that transitively imports a `"use client"` module that is reached during the export-types probe — e.g. a `forceClientPaths` entry pointing at a `node_modules` client-glue module, or a route graph that imports use-client glue eagerly.
- `npm run dev`. The export-types probe evaluates the worker entry before the scan completes; on a cold `.vite` this is frequently deterministic, and a browser tab reconnecting the instant the port binds makes it reliable.

## Suggested directions

1. Gate the worker-entry evaluation path (the `fetchModule`/`transformClientComponents`/`ssrLoadModule` chain the probe drives) on `scanPromise`, the same way the HTTP middleware is gated — so a `"use client"` module can't be evaluated before `clientFiles`/the vendor barrel are populated.
2. Fold the vendor-barrel content (or the sorted `clientFiles`/`serverFiles` membership) into the worker env's dep-optimizer cache key, so a prebundle built against an empty barrel is never `getConfigHash`-valid on the next boot.
3. Make `registerClientReference` tolerate an undefined / not-yet-registered SSR module — degrade to a retriable error instead of an uncaught `TypeError` that workerd can't recover from.

## Versions

rwsdk `1.2.12`, `@cloudflare/vite-plugin` `1.40.2`, vite `7.1.12`, workerd via miniflare 4.x.

chrsnymn:
Fixes #1231

When you have the dev server running and then add an import to a dependency that wasn't previously imported in that file, where that dependency contains a `"use client"` directive, the framework's internal directive lookup state — the vendor barrel — would not update, so the lookup would fail and the page would crash.

There were three interlocking reasons why this happened:

1. We were maintaining the vendor barrel inside `node_modules`. Because Vite's dependency optimizer treats `node_modules` as immutable, it didn't know the barrel had changed and never rebuilt it, so the dev server stayed on a stale version.
2. Even after making sure Vite rebuilds the barrel when it changes, that still isn't enough for an already-running dev server, because the module cache can still be stale.
3. Even if we solve the stale module problem, we still wouldn't know that the newly imported vendor package contains a `"use client"` directive — and this could even be true transitively, where a dependency re-exports from another package that has the directive.

The fix breaks the solution into parts that address each of those problems:

- We moved vendor barrel generation out of `node_modules` into a unique temp directory and serve their content in-memory via the existing esbuild barrel plugin. This prevents Vite from treating the barrels as immutable cached third-party files, so the optimizer rebuilds them when their content changes.
- For an already-running dev server, HMR now triggers a targeted directive re-scan starting from the changed worker module. The scan uses the same context-aware resolver as the initial startup scan, so it discovers newly imported directive modules with the correct client/server resolution conditions, including transitive imports through other packages.
- When the re-scan finds new directive files, we update the shared `clientFiles` / `serverFiles` sets, regenerate the vendor barrel content, and invalidate both the vendor barrel modules and the `virtual:use-client-lookup.js` / `virtual:use-server-lookup.js` modules in Vite's module graph. That forces the dev server to pick up the updated lookup state on the next request.

As a result, adding a `"use client"` dependency mid-session now works without restarting the dev server, and it also works when starting the dev server fresh even when an existing `.vite` cache is present.

Tests and docs are included: two new E2E tests in `playground/import-from-use-client/__tests__/` and updated architecture docs.

_**HMR performance caveat**: When a worker file imports a directive module that wasn't discovered in the initial scan, the HMR path currently awaits a full `runDirectivesScan`. This can delay the reload/refresh. It only happens for new dependency imports; already-known directives skip this branch. We're aware this is heavy and will monitor real-world impact. If it becomes a bottleneck, the next step is to cache scan results and/or make the sub-scan incremental._


justinvdm (approved):
🍰 !

PR #1239, opened 2026-06-24
fix: keep JSON and ?raw imports out of SSR bridge subgraph

### LINKED ISSUE #1221: Importing a JSON file crashes the dev server
A JSON file that is imported from a module ending up in **both** the client and server bundles crashes the rwsdk dev server while it scans for `"use client"` / `"use server"` directives. This happens both for a plain `import x from "foo.json"` and for `import x from "foo.json?raw"`.

This is the case discussed in the rwsdk Discord (June 2026), worked around in a real-world app with a custom `jsonModulePlugin`.

## Versions

- `rwsdk@1.3.0-canary.5`
- `vite@8.0.16`
- `@cloudflare/vite-plugin@1.39.0`
- `react@19.2.7` / `react-dom@19.2.7` / `react-server-dom-webpack@19.2.7`
- package manager: `pnpm`

## Reproduction:

https://github.com/valscion/rwsdk-reproductions/tree/main/rwsdk-json-import-repro

chrsnymn:
## Problem

Importing a JSON file from both a server component and a `"use client"` component crashed the dev server.

For plain `.json` imports, the SSR bridge pulled the JSON through Vite’s SSR environment, which emitted `__vite_ssr_exportName__` helpers. The Cloudflare worker runtime doesn’t provide those helpers, so the import failed at runtime.

Fixes #1221.

## Solution
Keep JSON and ?raw imports out of the SSR subgraph. When the bridge rewrites internal import callsites inside SSR-fetched code, it now leaves `.json` and `?raw` specifiers unchanged. They are resolved and transformed by the worker environment instead, where Vite and the Cloudflare Vite plugin handle them correctly.

## Changes
`sdk/src/vite/ssrBridgePlugin.mts`: skip the `virtual:rwsdk:ssr:` rewrite for `.json` and `?raw` imports.
`docs/architecture/ssrBridge.md`: document why non-JS assets are kept out of the SSR graph.
`playground/json-import/`: add a minimal repro with a server page, a `"use client"` component sharing JSON, a `?raw` route, and E2E tests.


justinvdm (approved):
🍰 !

PR #1241, opened 2026-06-25
Add navigation pending Suspense boundary

### LINKED ISSUE #1240: Expose client navigation pending/commit state for RSC navigations
## Problem

In built rwsdk apps under CPU pressure, client navigation can update the URL before the new RSC tree has committed. During that normal loading gap, the DOM still reflects the previous route/search/page state.

That is expected in principle: URL changes first, then the RSC request/commit updates the UI. The issue is that apps currently do not have a framework-level, composable way to say:

> “This subtree depends on the pending navigation state. While the pending navigation has not committed, suspend this subtree so React can show my normal `<Suspense fallback>`.”

Without that, server-backed search/filter/pagination UIs can keep showing stale rows as if they match the new URL.

## Observed behavior

In PRZM Playwright CI, especially when serving a built app (`vite preview`/CI-like environment) under load, we see cases like:

1. User navigates from page/search state B to A.
2. URL updates to A.
3. DOM remains showing B.
4. It may stay that way until another navigation or reload.

Examples from PRZM:

- Auction vehicle search:
  - URL has `?search=AVF-NDL`
  - table still shows unfiltered rows
- Auction vehicle empty search:
  - URL has `?search=ZZZZZZ-NO-MATCH`
  - “No vehicles match …” empty state never renders
- Auction vehicle pagination:
  - URL has `?page=2`
  - pagination text still says `Showing 1-50 of 60`

These are server-backed table states, so the app needs to hide stale rows and show loading UI while the RSC navigation is pending.

## Existing API limitation

Today `initClientNavigation({ onNavigate })` fires after history changes and before the RSC fetch, but there is no navigation-specific, composable pending/commit signal exposed to React components.

`initClient({ onHydrated })` exists, but it is global and not tied to a specific pending navigation target.

`navigate()` returns a promise, but it does not give components a Suspense-compatible “target URL has committed to the visible tree” primitive.

## Why plain Suspense is not enough today

`<Suspense>` is the right React primitive for the UI, but the app needs something inside the boundary that can suspend while rwsdk navigation is pending.

A server-backed client component often receives already-resolved server props:

```tsx
<AuctionVehicleTable
  vehicles={vehicles}
  vehicleSearch={vehicleSearch}
  vehiclePagination={vehiclePagination}
/>
```

Wrapping that in Suspense does not help by itself because the table can render old props immediately. There is no promise inside the table for Suspense to catch.

The missing piece is a framework-provided component/hook that understands rwsdk navigation lifecycle and can suspend until the pending RSC navigation commits.

## Proposed solution: `NavigationPending`

Expose a composable Suspense-aware primitive from `rwsdk/client`:

```tsx
import { Suspense } from 'react'
import { NavigationPending } from 'rwsdk/client'

<Suspense fallback={<AuctionVehicleTableSkeleton />}>
  <NavigationPending searchParams={['search', 'page']}>
    <AuctionVehicleTable
      vehicles={vehicles}
      vehicleSearch={vehicleSearch}
      vehiclePagination={vehiclePagination}
    />
  </NavigationPending>
</Suspense>
```

Conceptually:

- React `<Suspense>` owns the loading UI.
- rwsdk `NavigationPending` owns “should this subtree wait for the pending navigation commit?”
- If a relevant pending navigation is in flight, `NavigationPending` suspends on the navigation commit promise.
- When the pending RSC tree commits, the promise resolves and React renders the new subtree.

## API sketch

Default: suspend for any pending navigation.

```tsx
<Suspense fallback={<PageSkeleton />}>
  <NavigationPending>
    <Page />
  </NavigationPending>
</Suspense>
```

Targeted: suspend only when specific search params change.

```tsx
<Suspense fallback={<TableSkeleton />}>
  <NavigationPending searchParams={['search', 'page']}>
    <ResultsTable />
  </NavigationPending>
</Suspense>
```

More explicit watch config:

```tsx
<NavigationPending
  watch={{
    pathname: false,
    searchParams: ['search', 'page'],
    hash: false,
  }}
>
  <ResultsTable />
</NavigationPending>
```

Advanced escape hatch:

```tsx
<NavigationPending
  when={({ currentUrl, pendingUrl }) =>
    currentUrl.searchParams.get('tab') !==
    pendingUrl.searchParams.get('tab')
  }
>
  <TabPanel />
</NavigationPending>
```

## Framework internals needed

rwsdk would likely need to track navigation epochs:

1. navigation starts → assign `id`, record pending URL, create commit promise
2. URL/history updates
3. RSC fetch/stream begins
4. RSC payload for that navigation commits to the visible tree → resolve commit promise
5. newer navigation starts → cancel/supersede prior pending navigation
6. timeout/error → reject or resolve with a recovery state

Important: the promise should represent **commit to the visible React tree**, not just fetch completion.

## Optional timeout/recovery

It may also be useful for rwsdk to support a navigation commit timeout:

```ts
initClientNavigation({
  navigationTimeoutMs: 5000,
  onNavigationTimeout({ href }) {
    window.location.assign(href)
  },
})
```

This would prevent a Suspense fallback from hanging forever if the navigation is superseded, interrupted, or otherwise fails to commit.

## Prior art

Other frameworks expose route navigation pending state:

- React Router / Remix: `useNavigation()` with pending location/state
- Next App Router: route-segment `loading.tsx` and transition-based pending UI
- TanStack Router: pending components and pending timing options

React provides primitives like `Suspense`, `useTransition`, and `useDeferredValue`, but the URL-aware navigation lifecycle belongs in the router/framework.

## Notes

This is not asking rwsdk to choose app-specific loading UI. The app should own that through standard React `<Suspense fallback>`. The ask is for rwsdk to provide a composable primitive that suspends while relevant client navigation is pending, so apps do not show stale server-rendered data as if it matches the new URL.

peterp:
## Problem

Client-side navigation changes the URL before the new server-rendered React tree is on screen. During that short gap, a page can still show the old table rows or old pagination, even though the address bar already shows the new search or page. Apps had no built-in way to say, “hide this part until the matching navigation has really committed.”

## Solution

This adds `NavigationPending` and `useNavigationPending` to `rwsdk/client`. Apps can put `NavigationPending` inside their own `<Suspense fallback>` to show normal loading UI while a matching navigation is pending. The navigation runtime now tracks pending URLs and resolves them only when the matching RSC payload commits.

## Technical Summary

- Track one active navigation at a time with a small URL snapshot and commit promise. Newer navigations supersede older ones so old promises do not hang.
- Treat “committed” as the matching RSC payload reaching the visible React tree, not just the fetch finishing.
- Let apps scope pending UI by all navigations, selected search params, explicit URL parts, or a custom predicate.
- Ignore stale navigation responses when the browser has already moved to a newer URL, so an old response cannot update the tree late.
- Keep loading UI app-owned through React Suspense instead of adding a framework-specific fallback API.

Fixes #1240

### DIFF
--- a/docs/architecture/clientNavigationPending.md
+++ b/docs/architecture/clientNavigationPending.md
@@ -1,0 +1,174 @@
+---
+title: Client Navigation Pending Boundaries
+description: How RedwoodSDK exposes Suspense-aware pending state for client-side RSC navigations.
+---
+
+## Problem
+
+Client-side navigation updates the browser URL before the next React Server Component (RSC) payload has committed to the visible tree. During that gap, the old server-rendered subtree can still be visible even though the address bar already represents the new route, search params, or pagination state.
+
+That behavior is expected at the transport level: history changes first, then the client requests and renders the next RSC payload. The problem is that app code needs a composable way to say:
+
+> This subtree depends on the pending navigation. Hide it behind my normal Suspense fallback until the matching RSC payload has committed.
+
+A plain `<Suspense>` boundary is not enough when the subtree can render immediately from old, already-resolved server props. RedwoodSDK needs to provide the promise that represents the navigation commit.
+
+## Goals
+
+- Keep loading UI app-owned through React `<Suspense fallback>`.
+- Track pending navigations by target URL and commit promise.
+- Resolve the pending promise when the matching RSC payload commits to the visible React tree, not merely when fetch completes.
+- Prevent stale navigation responses from committing after the browser has moved to a newer URL.
+- Let apps scope pending UI by all navigations, selected search params, explicit URL parts, or a custom predicate.
+- Avoid changing the baseline progressive-enhancement model: links remain normal links without JavaScript.
+
+## Public API
+
+`NavigationPending` and `useNavigationPending` are exported from `rwsdk/client`.
+
+```tsx
+import { Suspense } from "react";
+import { NavigationPending } from "rwsdk/client";
+
+<Suspense fallback={<ResultsSkeleton />}>
+  <NavigationPending searchParams={["search", "page"]}>
+    <ResultsTable />
+  </NavigationPending>
+</Suspense>;
+```
+
+By default, `NavigationPending` suspends for any pending client navigation. Apps can narrow this with:
+
+- `searchParams`: shorthand for watching a list of search params.
+- `watch`: explicit URL parts (`pathname`, `searchParams`, `hash`).
+- `when`: custom predicate that receives copies of `currentUrl` and `pendingUrl`.
+
+If multiple options are supplied, the implementation evaluates them in this order: `when`, then `watch`, then `searchParams`, then the default "any pending navigation" behavior.
+
+## Main Pieces
+
+### `navigationState.ts`
+
+This module owns the external navigation store used by React components:
+
+- `currentUrl`: the URL represented by the React tree that has committed.
+- `pending`: the latest navigation that has updated history but has not committed.
+- `beginPendingNavigation(url)`: creates a pending navigation with an id, URL snapshot, and deferred promise.
+- `commitPendingNavigation(href)`: resolves and clears the pending navigation when a matching payload commits.
+- `abortPendingNavigation(id?)`: resolves and clears a pending navigation when it is superseded, redirected, or fails.
+- `subscribeNavigationState()` / `getNavigationSnapshot()`: power `useSyncExternalStore`.
+
+When a new navigation starts, any previous pending promise is resolved. This prevents old Suspense boundaries from hanging after they are superseded.
+
+### `navigation.ts`
+
+`initClientNavigation()` and `navigate()` begin the pending navigation after history has been updated and before the RSC request is made.
+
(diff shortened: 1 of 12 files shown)


peterp:
Follow-up notes from review (not blockers for this PR):

- We may eventually want per-navigation request IDs/epochs so rapid navigations to the exact same URL can be distinguished more explicitly than URL matching alone.
- Aborting superseded in-flight RSC fetches with `AbortController` could save resources, but correctness should still rely on the current URL/commit guard because aborts are best-effort.
- The legacy `rwsdk/realtime` transport has a similar stale-response metadata risk, but realtime is being deprecated and is separate from `use-synced-state`, so I did not treat it as blocking here.

I pushed a small fix for the hash-only case so `#hash` changes do not make a valid RSC navigation response look stale or leave `NavigationPending` suspended forever.

peterp:
Created follow-up sub-issues under #1240:

- #1242 Track client navigation commits with request ids
- #1243 Abort superseded in-flight RSC navigation fetches
- #1244 Decide stale navigation handling for deprecated realtime transport

justinvdm (approved):
Super cool!

PR #1247, opened 2026-06-25
Export NavigationPending from client SSR entry

### LINKED ISSUE #1246: Export NavigationPending from rwsdk/client SSR entry
## Problem

`NavigationPending` is exported from `rwsdk/client` for the browser build and from the public TypeScript types, but it is missing from the `workerd` / SSR client entry.

That means app code can typecheck with a normal static import:

```ts
import { NavigationPending } from 'rwsdk/client'
```

…but a production build can fail when the worker/SSR bundle resolves `rwsdk/client` to `dist/runtime/entries/clientSSR.js`:

```txt
"NavigationPending" is not exported by ".../rwsdk/dist/runtime/entries/clientSSR.js"
```

We hit this in PRZM while using the new navigation pending support from #1241 / v1.4.0. The feature works in the browser, but apps cannot safely use the public import without a workaround.

## Solution

Export the navigation pending API from the `workerd` / SSR client entry too.

At minimum, `rwsdk/client` should expose the same named exports across its browser, type, and worker/SSR conditions:

- `NavigationPending`
- `useNavigationPending`
- related public types are already available from the public type entry

For SSR/workerd rendering, `NavigationPending` can be a safe no-op wrapper that returns its children. There is no browser client navigation pending during SSR, so it should not suspend there.

## Technical Summary

- Keep the public `rwsdk/client` API consistent across package export conditions. A valid import should not depend on which bundle Vite is currently building.
- The SSR/workerd implementation does not need to track navigation state. The important guarantee is that SSR can render the component tree without throwing or hiding content.
- `useNavigationPending` can return a non-pending snapshot in SSR/workerd, or be wired to the shared implementation if that is safe in this environment.
- Add coverage for statically importing `NavigationPending` from a client component that is included in a worker/SSR build, because TypeScript alone did not catch this mismatch.

Related PRZM workaround: https://github.com/goprzm/przm/pull/1780

peterp:
## Problem

`NavigationPending` is part of the public `rwsdk/client` API, and the TypeScript types say it is available. But the worker/SSR client entry did not export it. That meant an app could typecheck, then fail during a production build when the worker bundle tried to import it.

## Solution

The worker/SSR client entry now exports `NavigationPending` and `useNavigationPending`. In that environment there is no browser navigation in progress, so the boundary simply renders its children and the hook returns a non-pending snapshot.

## Technical Summary

- Keep `rwsdk/client` named exports consistent across browser and worker/SSR package conditions.
- Use no-op SSR behavior instead of importing browser navigation state into the worker/SSR entry. This keeps SSR rendering simple and safe.
- Add direct coverage for the worker/SSR entry exports so a public type/export mismatch is caught before release.
- Fixes #1246.

### DIFF
--- a/sdk/src/runtime/entries/clientSSR.ts
+++ b/sdk/src/runtime/entries/clientSSR.ts
@@ -1,6 +1,26 @@
 import "./types/ssr";
+
+import type {
+  NavigationPendingOptions,
+  NavigationPendingProps,
+} from "../client/navigationPending.js";
+import type { NavigationSnapshot } from "../client/navigationState.js";
+
 export * from "../lib/streams/consumeEventStream";
 
 export const navigate = () => {
   /* stub */
 };
+
+export function useNavigationPending(
+  _options: NavigationPendingOptions = {},
+): NavigationSnapshot {
+  return {
+    currentUrl: new URL("http://localhost/"),
+    pending: null,
+  };
+}
+
+export function NavigationPending({ children }: NavigationPendingProps) {
+  return children ?? null;
+}
--- a/sdk/src/runtime/entries/clientSSR.test.ts
+++ b/sdk/src/runtime/entries/clientSSR.test.ts
@@ -1,0 +1,31 @@
+import React from "react";
+import { renderToString } from "react-dom/server";
+import { describe, expect, it } from "vitest";
+
+import { NavigationPending, useNavigationPending } from "./clientSSR";
+
+function NavigationSnapshotProbe() {
+  const snapshot = useNavigationPending();
+
+  return React.createElement(
+    "span",
+    null,
+    snapshot.pending === null ? "not-pending" : "pending",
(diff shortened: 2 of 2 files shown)


sdk/src/runtime/entries/clientSSR.ts
```diff
@@ -1,6 +1,26 @@
...
   /* stub */
 };
+
+export function useNavigationPending(
+  _options: NavigationPendingOptions = {},
+): NavigationSnapshot {
+  return {
+    currentUrl: new URL("http://localhost/"),
+    pending: null,
+  };
```
peterp, line 21: This feels bad and wrong.

justinvdm: I know what you mean. At same time, if the same api's have to get run for ssr as for client, I think it might be the only way.

PR #1249, opened 2026-06-26
chore: upgrade Vite to 8.1.0

justinvdm:
## What

Upgrades Vite across the monorepo from `~8.0.9` to `~8.1.0` so the SDK and all playgrounds run on the latest Vite 8 release.

## Problem

After bumping the Vite pins, the Playground E2E deploy job surfaced a production runtime error in `playground/database-do`:

```
Uncaught ReferenceError: freeze is not defined
```

The same error reproduced locally with `vite build && vite preview`.

## Why

Vite 8 uses Rolldown for production builds. For packages marked `sideEffects: false`, Rolldown can emit lazy ESM wrappers that fail to rebind named imports to their minified top-level bindings. kysely's `freeze` helper was renamed correctly in most places, but the wrapper around `CompiledQuery.raw` still called `freeze(...)` directly, leaving it undefined at runtime.

## Solution

- Updated the Vite pin in every workspace `package.json` from `~8.0.9` to `~8.1.0`.
- Ran `pnpm install` to update the lockfile so all Vite resolutions point to `8.1.0`.
- Disabled `experimental.lazyBarrel` in `playground/database-do/vite.config.mts` for the worker environment. This is a playground-local workaround scoped to the affected example until the upstream Rolldown bug is fixed.

This workaround can be removed once the upstream Rolldown issue is fixed.

## Verification

- Confirmed no `~8.0.9` pins or `vite@8.0.9` lockfile entries remain.
- `playground/database-do` builds and previews without the runtime error.
- Rebuilt all playground worker bundles and searched for bare `freeze` references; none were found in `database-do` or elsewhere.


justinvdm:
Closing this PR for now. The Vite 8.1.0 upgrade surfaced a Rolldown production bundling issue (lazy ESM wrappers can leave named imports unbound, e.g. `freeze is not defined` in kysely). The targeted package-specific workaround we tried is not the right shape, so we are going to figure out a general fix before re-opening.

justinvdm:
Still failing CI with the kysely upgrade alone. The Rolldown lazy ESM wrapper bug remains, so we need a general fix rather than package-specific workarounds. Closing until we have that.

PR #1256, opened 2026-07-01
fix(vite): honor optimizeDeps.exclude for directive node_modules files

### LINKED ISSUE #1250: Allow opting specific node_modules "use client" modules out of the vendor barrel (serve from source)
### Problem

In dev, RedwoodSDK routes every `node_modules` `"use client"` (and `"use server"`) module into the prebundled **vendor barrel**: `generateLookupMap` (`createDirectiveLookupPlugin`) emits an `import(VENDOR_CLIENT_BARREL_EXPORT_PATH)` for any file whose path contains `node_modules`, and `generateVendorBarrelContent` (`directiveModulesDevPlugin`) `import * as`-es them all. That's a sensible default for typical pre-built libraries.

But there's no opt-out for a **host framework that ships an ESM-source `"use client"` component library** intended to be transformed by the host's own Vite pipeline. Two consequences:

1. **The host's transform never runs on those modules.** If a library's components carry per-component compiled styling (a CSS-in-JS transform that emits a component's CSS only when that module loads), routing them through the prebundled barrel skips that transform entirely.
2. **No tree-shaking.** The vendor barrel imports every client leaf, so the *whole* library compiles regardless of which components the app actually uses.

`forceClientPaths` only controls *what* is treated as a client file — not *how* it's served. The source-vs-barrel decision is a hard `file.includes("node_modules")` check with no escape hatch.

### Proposal

A `forceSourcePaths` option on `redwoodPlugin`, mirroring `forceClientPaths` and resolved through the same `resolveForcedPaths` pipeline. Matched client/server files take the **existing** `() => import("${file}")` source branch in `generateLookupMap` even under `node_modules`, and are excluded from the vendor barrel. One supporting fix is required: strip the optimizer `?v=<hash>` query before `normalizeModulePath` in `directivesPlugin`, so a source-served module referenced through an optimized chunk registers its client-reference id under the plain path that matches the lookup-map key.

```ts
redwoodPlugin({ forceSourcePaths: ["node_modules/<pkg>/**/*.{js,jsx,ts,tsx}"] })
```

Default behavior is unchanged when the option is unset.

### Motivation

I'm building a meta-framework on RedwoodSDK whose component library ships per-component compiled styling. Source-serving lets the styling transform run only on the components that actually load (only-used CSS) and keeps dev/prod parity — neither is achievable while the components are frozen into the vendor barrel.

I have a PR ready that implements this with tests, and I'm happy to adjust the API/naming to fit your preferences.

chrsnymn:
Fixes redwoodjs/sdk#1250.

When a package is listed in `optimizeDeps.exclude`, Vite opts it out of pre-bundling. RedwoodSDK was not honoring that for directive-bearing `node_modules` files: they were still routed through the pre-bundled **vendor barrel**, which bypasses the host's Vite transform pipeline. That broke libraries that rely on user plugins to inject CSS imports, rewrite JSX, add runtime markers, etc.

This PR moves excluded `node_modules` directive files into the **app barrel** so they run through the host Vite pipeline.

### Changes

- Added `sdk/src/vite/resolveOptimizeDepsExcludes.mts` to resolve `optimizeDeps.exclude` patterns into absolute package roots. Supports bare packages, scoped packages, subpaths, relative/absolute paths, and symlinked packages. Root-level and per-environment excludes are unioned.
- Updated `sdk/src/vite/directiveModulesDevPlugin.mts` so excluded `node_modules` files are included in the app barrel instead of the vendor barrel.
- Updated `sdk/src/vite/createDirectiveLookupPlugin.mts` to source-import excluded `node_modules` files in dev rather than routing them through the vendor barrel.
- Stripped optimizer `?v=<hash>` query strings in `sdk/src/vite/directivesPlugin.mts` before normalizing module IDs, so source-served modules match their lookup keys.
- Added/expanded unit tests for the resolver, barrel generation, and lookup map.
- Added `playground/vendor-barrel-source-repro` with an E2E test that verifies host transforms run in dev.
- Updated `docs/architecture/devServerDependencyOptimization.md` to document the new `optimizeDeps.exclude` behavior.

### Verification

SDK unit tests for the affected modules:

```sh
cd sdk
pnpm vitest --run \
  src/vite/resolveOptimizeDepsExcludes.test.mts \
  src/vite/directiveModulesDevPlugin.test.mts \
  src/vite/createDirectiveLookupPlugin.test.mts \
  src/vite/directivesPlugin.test.mts
  ```
  
  Result: **42 tests passed**

Repro E2E (dev only — deployment requires Cloudflare auth locally):
```
RWSDK_SKIP_DEPLOY=1 pnpm test:e2e -- vendor-barrel-source-repro/__tests__/e2e.test.mts
```
Result: **passed**

The full test suite was not completed locally because deployment E2Es require Cloudflare authentication.

### CODE REVIEW

**Comment by justinvdm** [comment]
Closing and re-opening for CI to recognize added label

**Review by justinvdm — APPROVED** [review]
🍰 !

**Comment by Zn4rK** [comment]
Thanks for landing this in 1.5.7 — the `optimizeDeps.exclude` → app-barrel approach works for source-serving an excluded `node_modules` `"use client"` package (I'm on 1.5.7 and the direct case is fixed).

There's one remaining gap that I believe this PR intended to cover. When a source-served (excluded) `"use client"` component is reached **transitively through a prebundled chunk** — including RedwoodSDK's own prebundled worker `router` — Vite bakes the optimizer `?v=<hash>` query into that prebundled chunk's rewritten import of the component. The directive lookup registers the client reference under the plain normalized path, so the worker/SSR lookup for the `?v=`-suffixed request misses:

```
(ssr) No module found for '…/node_modules/<pkg>/<component>.js?v=908de5ea' in module lookup for "use client" directive
```

The hash varies per optimize run (I've seen `908de5ea` and `b7b4c446` for the same module), which confirms it's the optimizer browserHash frozen into the prebundled importer's import — not something a downstream `resolveId` hook can intercept (by the time the module runner requests it, the query is already baked into the prebundled chunk).

The PR description says *"Stripped optimizer `?v=<hash>` query strings in `directivesPlugin.mts` before normalizing module IDs"*, but the released `dist/vite/directivesPlugin.mjs` normalizes the raw id in its `transform`:

```js
const normalizedId = normalizeModulePath(id, projectRootDir);
```

with no query stripping there (nor in `normalizeModulePath`). Stripping the query first appears to fix it:

```js
const normalizedId = normalizeModulePath(id.split("?")[0], projectRootDir);
```

so the registered client-reference id matches the plain lookup-map key regardless of how the module was reached.

**Minimal repro shape:** a `node_modules` package exporting a `"use client"` component, listed in `optimizeDeps.exclude`, imported **transitively through a prebundled dependency** (or any optimized chunk) rather than directly from app source. The direct-from-source path works; the through-a-prebundled-chunk path fails the use-client lookup as above.

Happy to open a small follow-up PR with the one-line strip if that's useful. (Refs #1250.)

chrsnymn:
Hi, we can repro, we're busy making sure we can fully understand what the issue is. Will get back to you on this.

justinvdm:
Sorry for the long delay, we understand the issue now, repro+fix PR incoming.

justinvdm:
@Zn4rK fix should be released in https://github.com/redwoodjs/sdk/releases/tag/v1.5.9 :)

PR #1258, opened 2026-07-02
fix(render): avoid false-positive tag split on chunk boundary during hoisting

### LINKED ISSUE #1257: `<title>/<meta>` hoisting appears to fail under a stream chunk-boundary edge case in `stitchDocumentAndAppStreams.js`
I ran into an intermittent production issue where `<title>` occasionally rendered unclosed, which in turn caused the browser (Safari and Chrome) to swallow the rest of the document, including the hydration root. This broke hydration entirely. 

Screen Capture of broken title tag in Safari

The issue appeared only showed up on some pages and not others, and toggling unrelated content on a page could make the issue appear or disappear. This changed the overall length of the page, but the specific content itself didn't matter. The content was all valid, no errant HTML tags or unsanitized HTML.

`splitStreamOnFirstNonHoistedTag` in `stitchDocumentAndAppStreams.js` (added in [#885](https://github.com/redwoodjs/sdk/pull/885) to fix [#872](https://github.com/redwoodjs/sdk/issues/872)) buffers the app HTML stream incrementally and tests this regex against the growing buffer to find the first tag that isn't title/meta/link/style/base:

```js
const nonHoistedTagPattern = /<(?!(?:\/)?(?:title|meta|link|style|base)[\s>\/])(?![!?])/I;
```

I believe that if a chunk boundary happens to land in the middle of one of those tag names,  like if the buffer ends at `<t` or an unclosed `<title>Foo</title`, the negative lookahead can't resolve against a partial tag name and then produces a false-positive match:

```js
const pattern = /<(?!(?:\/)?(?:title|meta|link|style|base)[\s>\/])(?![!?])/i;
"<t".match(pattern);                 // matches "<" — looks like a false positive to us
"<title>Foo</title".match(pattern);  // matches "<" (the closing tag, missing ">") — same issue
```
So it looks like the right accidental truncation would sever the `<title>` tag in the output causing the hydration bugs I saw.

I think maybe this is related to `writeChunk` in react-server-dom's server renderer. 2048 byte rolling buffer which gets flushed when the buffer overflows.

I was not able to reproduce this locally in dev, only in production. I assume the dev server does not preserve the same incremental streaming boundaries as production Workers.

Working with Claude Code, I created a patch which seems to mitigate the issue:

```patch
diff --git a/dist/runtime/lib/stitchDocumentAndAppStreams.js b/dist/runtime/lib/stitchDocumentAndAppStreams.js
index 989785e317401d11b3fd34a74b5391f7dab905cb..f6278e298d8201f5deee525c93ae331ef7adad33 100644
--- a/dist/runtime/lib/stitchDocumentAndAppStreams.js
+++ b/dist/runtime/lib/stitchDocumentAndAppStreams.js
@@ -19,6 +19,24 @@ function splitStreamOnFirstNonHoistedTag(sourceStream) {
     const decoder = new TextDecoder();
     const encoder = new TextEncoder();
     const nonHoistedTagPattern = /<(?!(?:\/)?(?:title|meta|link|style|base)[\s>\/])(?![!?])/i;
+    // Longest hoistable name is 5 chars (title/style); a match candidate at
+    // index i needs buffer[i..] to have at least 1 (optional "/") + 5 (name)
+    // + 1 (terminating char) = 7 chars of context past "<" before it can be
+    // trusted as final. Otherwise the tag name may be mid-stream-truncated,
+    // causing a false-positive match on e.g. "<t" and severing <title>.
+    const MIN_TRAILING_CONTEXT = 7;
+    function findTrustedMatch(buf, streamDone) {
+        const match = buf.match(nonHoistedTagPattern);
+        if (!match || typeof match.index !== "number")
+            return null;
+        if (streamDone)
+            return match; // no more data coming; trust it
+        // buf.length - match.index counts the "<" itself, so subtract 1 to
+        // get the number of confirmed characters *after* it.
+        if (buf.length - match.index - 1 < MIN_TRAILING_CONTEXT)
+            return null; // ambiguous, wait for more data
+        return match;
+    }
     let sourceReader;
     let appBodyController = null;
     let buffer = "";
@@ -35,7 +53,7 @@ function splitStreamOnFirstNonHoistedTag(sourceStream) {
                     const { done, value } = await sourceReader.read();
                     if (done) {
                         if (buffer) {
-                            const match = buffer.match(nonHoistedTagPattern);
+                            const match = findTrustedMatch(buffer, true);
                             if (match && typeof match.index === "number") {
                                 const hoistedPart = buffer.slice(0, match.index);
                                 controller.enqueue(encoder.encode(hoistedPart));
@@ -52,7 +70,7 @@ function splitStreamOnFirstNonHoistedTag(sourceStream) {
                         return;
                     }
                     buffer += decoder.decode(value, { stream: true });
-                    const match = buffer.match(nonHoistedTagPattern);
+                    const match = findTrustedMatch(buffer, false);
                     if (match && typeof match.index === "number") {
                         const hoistedPart = buffer.slice(0, match.index);
                         const appPart = buffer.slice(match.index);
@@ -76,9 +94,19 @@ function splitStreamOnFirstNonHoistedTag(sourceStream) {
                     }
                     else {
                         const flushIndex = buffer.lastIndexOf("\n");
-                        if (flushIndex !== -1) {
-                            controller.enqueue(encoder.encode(buffer.slice(0, flushIndex + 1)));
-                            buffer = buffer.slice(flushIndex + 1);
+                        // Never flush past a point that could still be part of
+                        // an ambiguous, not-yet-trusted candidate tag. Only
+                        // flush content strictly before the last unresolved
+                        // "<" in the buffer, so a truncated tag name can never
+                        // be split across a flush boundary.
+                        const earliestOpenBracket = buffer.lastIndexOf("<");
+                        const safeFlushLimit = earliestOpenBracket === -1
+                            ? buffer.length
+                            : earliestOpenBracket;
+                        const effectiveFlushIndex = Math.min(flushIndex, safeFlushLimit - 1);
+                        if (flushIndex !== -1 && effectiveFlushIndex >= 0) {
+                            controller.enqueue(encoder.encode(buffer.slice(0, effectiveFlushIndex + 1)));
+                            buffer = buffer.slice(effectiveFlushIndex + 1);
                         }
                         await pump();
                     }
```

chrsnymn:
`<title>`, `<meta>`, `<link>`, `<style>`, and `<base>` tags rendered by the React app are hoisted into `<head>` by buffering the app stream and scanning for the first non-hoisted tag. The scanner used a regex that could produce false-positive matches when a stream chunk boundary landed in the middle of a tag name.

For example, if a chunk ended at `<t `or at `</title` (before the closing >), the regex would match the < as the start of a non-hoisted tag and split the stream there. This either dropped the title into the body or, worse, left an unclosed `<title>` in `<head>`, which caused the browser to swallow the rest of the document and broke hydration.

### Changes
Added a findTrustedMatch helper in splitStreamOnFirstNonHoistedTag that only trusts a regex match when there is enough trailing context (7 characters past <) to confirm the full tag name, or when the stream has ended.
Tightened the flush logic so we never flush content past the last unresolved < in the buffer, preventing a truncated tag from being split across a flush boundary.

### Verification
Added regression unit tests in sdk/src/runtime/lib/stitchDocumentAndAppStreams.test.ts covering:
- Closing </title> split before >
- <title tag name split as <t / itle>
- pnpm test src/runtime/lib/stitchDocumentAndAppStreams.test.ts → 22 passed
-  pnpm test → 610 passed, 1 skipped
- pnpm --filter rwsdk build succeeds

No playground example was added because this is a core stream-transformation bug that only reproduces under specific production chunk boundaries; the unit tests directly exercise the affected function with controlled chunk boundaries.

Fixes #1257

### DIFF
--- a/sdk/src/runtime/lib/stitchDocumentAndAppStreams.test.ts
+++ b/sdk/src/runtime/lib/stitchDocumentAndAppStreams.test.ts
@@ -229,6 +229,90 @@ describe("stitchDocumentAndAppStreams", () => {
 
       expect(result.trim().startsWith(`<!DOCTYPE html>`)).toBe(true);
     });
+
+    it("does not split an unclosed title tag across a chunk boundary", async () => {
+      const outerHtml = `<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8" />
+</head>
+<body>
+  ${startMarker}
+  <script src="/client.js"></script>
+</body>
+</html>`;
+
+      // The ">" of the closing </title> arrives in the next chunk. A naive
+      // regex scan of the first chunk can falsely treat the "</title" as a
+      // non-hoisted tag boundary, leaving the <title> unclosed in <head>.
+      const innerHtmlChunks = [
+        `<title>Page Title</title`,
+        `><div>App content</div>${endMarker}`,
+      ];
+
+      const result = await streamToString(
+        stitchDocumentAndAppStreams(
+          stringToStream(outerHtml),
+          createChunkedStream(innerHtmlChunks),
+          startMarker,
+          endMarker,
+        ),
+      );
+
+      expect(result).toContain(`<title>Page Title</title>`);
+      expect(result).toMatch(
+        /<head>[\s\S]*<title>Page Title<\/title>[\s\S]*<\/head>/,
+      );
+      expect(result).toContain(`<div>App content</div>`);
+
+      const titleIndex = result.indexOf(`<title>Page Title</title>`);
+      const headCloseIndex = result.indexOf(`</head>`);
+      const bodyIndex = result.indexOf(`<body>`);
+
+      expect(titleIndex).toBeLessThan(headCloseIndex);
+      expect(titleIndex).toBeLessThan(bodyIndex);
+    });
+
+    it("does not truncate a hoisted tag name across a chunk boundary", async () => {
+      const outerHtml = `<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8" />
+</head>
+<body>
+  ${startMarker}
+  <script src="/client.js"></script>
+</body>
+</html>`;
+
+      // The "<title" tag name is split between chunks. A naive regex scan of
+      // the first chunk (ending at "<t") can falsely treat that "<" as a
+      // non-hoisted tag boundary and fail to hoist the title.
+      const innerHtmlChunks = [
+        `<t`,
+        `itle>Page Title</title><div>App content</div>${endMarker}`,
+      ];
+
+      const result = await streamToString(
+        stitchDocumentAndAppStreams(
+          stringToStream(outerHtml),
+          createChunkedStream(innerHtmlChunks),
+          startMarker,
+          endMarker,
+        ),
+      );
+
+      expect(result).toContain(`<title>Page Title</title>`);
+      expect(result).toMatch(
+        /<head>[\s\S]*<title>Page Title<\/title>[\s\S]*<\/head>/,
+      );
+      expect(result).toContain(`<div>App content</div>`);
+
+      const titleIndex = result.indexOf(`<title>Page Title</title>`);
+      const headCloseIndex = result.indexOf(`</head>`);
+
+      expect(titleIndex).toBeLessThan(headCloseIndex);
+    });
   });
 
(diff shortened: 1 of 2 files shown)


justinvdm (approved):
🍰 !

Issue #1259, opened 2026-07-02
Dev cold start crashes with MISSING_EXPORT registerClientReference when an auxiliary worker is combined with any node_modules "use client" dependency


dethstrobe:
## Summary

In dev, a cold start (`node_modules/.vite` absent) crashes during dependency optimization when a project combines:

1. **any auxiliary worker** (`cloudflare({ auxiliaryWorkers: [...] })` from `@cloudflare/vite-plugin`), and
2. **any `node_modules` dependency whose dist files carry `"use client"`** (`@radix-ui/*`, `sonner`, `next-themes`, `swr`, `@ai-sdk/react`, ...).

Every optimizer bundle for the auxiliary worker's environment fails with:

```
[MISSING_EXPORT] "registerClientReference" is not exported by
"node_modules/.pnpm/rwsdk@1.5.5_.../rwsdk/dist/runtime/entries/react-server-only.js".
   ╭─[ node_modules/.pnpm/sonner@2.0.7_.../sonner/dist/index.mjs:2:10 ]
   │
 2 │ import { registerClientReference } from "rwsdk/worker";
   │          ───────────┬───────────
   │                     ╰───────────── Missing export
```

Either ingredient alone is fine; together they break every cold start. Warm optimizer caches mask the bug, so it typically surfaces later as "dev suddenly won't boot" (e.g. in CI, after `pnpm clean`, or on a coworker's machine) rather than at the commit that introduced the second ingredient.

## Reproduction

https://github.com/dethstrobe/applywize2/tree/registerClientReference_bug — the repro is a 3-file diff on an ordinary rwsdk app that already used Radix UI ([commit](https://github.com/dethstrobe/applywize2/commit/4f8d73b)): bump `rwsdk` to 1.5.5 / `vite` to 8.1.0, add a 5-line stub worker, and register it as an auxiliary worker.

```sh
git clone -b registerClientReference_bug https://github.com/dethstrobe/applywize2
cd applywize2 && pnpm install
rm -rf node_modules/.vite && pnpm dev   # crashes as above
```

Remove the `auxiliaryWorkers` line in `vite.config.mts`, and the same cold start boots cleanly.

Results observed (all cold starts):

| rwsdk | vite | node_modules "use client" dep | aux worker | result |
|---|---|---|---|---|
| 1.4.0 | 7.3.6 | yes (radix, swr) | no | ✅ boots |
[...]

PR #1261, opened 2026-07-08
fix(vite): scope directive and barrel setup to SDK environments

chrsnymn:
**Fixes**: #1259

### Summary
In dev, combining an auxiliary worker (`@cloudflare/vite-plugin auxiliaryWorkers`) with any `node_modules` dependency that ships `"use client"` files crashes the cold start with:

`[MISSING_EXPORT] "registerClientReference" is not exported by "rwsdk/worker"`
The SDK only owns three Vite environments (`client`, `ssr`, `worker`), but the Cloudflare plugin creates extra environments for auxiliary workers. The SDK's directive/barrel plugins were configuring every environment, so the auxiliary worker's optimizer tried to bundle the vendor client barrel, got rewritten to import `registerClientReference` from `rwsdk/worker`, and failed because the auxiliary environment lacks the `react-server` resolve condition.

### Changes
- Added `SDK_ENVIRONMENT_NAMES = ["client", "ssr", "worker"]` in `sdk/src/vite/constants.mts`.
- Scoped `directiveModulesDevPlugin` to only inject vendor/app barrels into SDK-owned environments.
- Scoped `directivesPlugin` to only register the `rsc-directives-transform` optimizer plugin for SDK-owned environments.
- Added unit tests for both plugins verifying auxiliary/custom environments are skipped.
- Added a reproduction playground: `playground/aux-worker-registerclientreference` (uses `sonner`, a real `"use client"` package, plus an auxiliary worker).
- Updated architecture docs in `directiveTransforms.md` and `devServerDependencyOptimization.md`.

### Verification
```
# Unit tests
cd sdk && pnpm test

# Reproduction playground E2E (deploy test skipped locally)
RWSDK_SKIP_DEPLOY=1 pnpm test:e2e -- playground/aux-worker-registerclientreference/__tests__/e2e.test.mts
```
Both pass.

> [!NOTE]
> Auxiliary workers are intentionally left as plain Workers. They do not receive RSC transforms or vendor/app barrels. If a user needs RSC behavior in an auxiliary worker, that is outside the scope of this fix.


justinvdm (approved):
🍰 makes sense!

PR #1263, opened 2026-07-16
Fix HMR-triggered runDirectiveScan to only happen on import change

chrsnymn:
Handles https://github.com/goprzm/sdk/pull/8

Gates the HMR-triggered directive sub-scan on import changes so that body-only edits to worker files no longer run the expensive `runDirectivesScan` on every save.

### Problem
In the dev server, `runDirectivesScan` was being invoked on every HMR update to any file already in the worker module graph. This caused the esbuild process to accumulate CPU and memory usage over a long dev session.

Two issues led to this:

1. `isWorkerUpdate` was always true because `Boolean(modules)` is true for any array.
2. There was no check for whether the changed file's imports had actually changed.

### Changes
`sdk/src/vite/miniflareHMRPlugin.mts`
- Fixes isWorkerUpdate to modules.length > 0.
- Adds a module-level `importSignatureCache`.
- Adds getImportSignature(id, code) using the existing findImportSpecifiers utility.
  - Skips runDirectivesScan when the file's import signature matches the cached value.
  - Adds debug logs for "signature changed / unchanged".

`sdk/src/vite/miniflareHMRPlugin.test.mts`
- Adds unit tests for getImportSignature.
- Adds repro tests that exercise the actual hotUpdate hook:
  - first update runs the scan
  - body-only update skips the scan
  - import change runs the scan again

### Verification
Run from the `sdk/` directory:
```
pnpm build
pnpm vitest run src/vite
```
Results on this branch:

pnpm build ✅
pnpm vitest run src/vite → *257 passed*
I also manually verified the behavior in playground/directives by running:
```
cd playground/directives
DEBUG=rwsdk:vite:hmr-plugin pnpm dev
```
and editing `src/worker.tsx`:

- body-only edits after the first one logged hmr: import signature unchanged, skipping directive scan
- import changes logged hmr: import signature changed, running directive scan

### Notes
This fix reduces how often `runDirectivesScan` runs, which addresses the reported CPU/memory growth symptom. A separate investigation is still needed to determine whether runDirectivesScan itself leaks memory per invocation.
The repro tests use `vi.mock()` for `runDirectivesScan` and `invalidateModule` because the plugin is not currently structured for dependency injection. The mocking is limited to the HMR gating tests.


justinvdm:
🍰 !

PR #1271, opened 2026-08-02
feat(client-navigation): add first-class full-reload opt-out for navigation

### LINKED ISSUE #1267: First-class full-reload API for navigation across client entry points
## Problem

An app can serve different sections from different client entry points, each booted by its own Document — for example, `/` from one entry and `/admin` from another. On client navigation, the server renders only the target route's page element; the Document is consumed solely for full HTML loads. So a soft navigation across entries applies the new page inside the *previously loaded* Document shell, under the *previous entry's* client program.

When the two entries use materially different Documents and client programs, the target page renders without what its own Document and program provide:

- **Different stylesheets** — the two sheets may share design-token names with different values, and one sheet may define tokens the other lacks. Cross-applied content renders mis-styled (wrong accent, ring, and background values) or partially unstyled (utilities referencing tokens the loaded sheet never defines are dropped).
- **Different globals** — anything the target's Document installs (analytics scripts, fonts, `<title>`, meta, html/body attributes) never arrives; the previous Document's globals remain in force.
- **Different client programs** — each entry boots its own client logic (monitoring init, scroll management, `initClientNavigation` options). After crossing the boundary, the previous entry's logic still governs the new page.

For apps in this situation, a full document reload at the boundary is the correct behavior: the browser must load the target's Document and client program fresh. This is not a module-loading limitation — a minimal two-entry reproduction with symmetrical shells navigates across entries cleanly. The failure mode appears when the two shells differ materially.

## Today: no API for it, only a side effect

rwsdk currently offers no first-class way to request a full reload for a given navigation. Apps work around this by setting `target="_top"` on boundary-crossing anchors — which functions only because `validateClickEvent` in `sdk/src/runtime/client/navigation.ts` declines links whose `target` is set and is not `_self`, so the click falls through to a normal browser navigation. That is an undocumented side effect of an unrelated check, and it covers only anchor clicks.

## Solution ideas

1. **An opt-out attribute on anchors** — `<a href="/admin" data-reload>`. Declarative and colocated with the link; precedent: React Router's `<Link reloadDocument>`, Turbo's `data-turbo="false"`. Covers anchor clicks only.
2. **A global rules callback on `initClientNavigation`** — e.g. `shouldIntercept({ url, element })`. Expresses structural rules ("everything under `/admin`") and is the only shape that could also cover `navigate()` and `popstate` (back/forward across a boundary, which an attribute cannot reach).
3. **An option on `navigate()`** — `navigate(href, { reload: true })`. Convenience; `navigate()` already performs a real document navigation when client navigation was never initialized.

Design notes: the interceptor only controls `preventDefault()`, so the honest semantics are "let the browser perform its default navigation"; and if rules ever exclude paths, the prefetch path (`<link rel="x-prefetch">`) needs a matching decision.

Worth documenting alongside whichever shape lands: apps that *want* soft navigation across such sections can alternatively design their entries to share one stylesheet/token set and one client program, making the shells navigation-invariant.

chrsnymn:
**Closes #1267**

## Problem

RedwoodSDK turns clicks on internal links into soft (client-side) navigations by default. That works inside one section of an app, but it breaks when you cross a boundary between two sections that use different Documents — for example, a public site at `/` and an admin area at `/admin`.

This PR adds a first-class way to say "do a full browser reload for this navigation" so the target section's Document, stylesheets, and client code are loaded fresh.

### 1. `data-reload` on an anchor

```tsx
<a href="/admin" data-reload>Admin</a>
```
Clicking this link forces a full browser reload instead of a soft navigation.

### 2. `shouldIntercept` callback on `initClientNavigation`
```tsx
initClientNavigation({
  shouldIntercept({ toUrl, fromUrl }) {
    const isAdmin = (url: URL) => url.pathname.startsWith("/admin");
    return isAdmin(toUrl) === isAdmin(fromUrl);
  },
});
```
Return `true` for a soft navigation, `false` for a full reload. This callback runs for:

- anchor clicks
- `navigate()` calls
- back/forward (`popstate`) navigation
### 3. Prefetch stays in sync
The prefetch helper also skips any `<link rel="x-prefetch">` link that would result in a full reload, so we don't waste time warming a cache entry that will never be used.

## Files changed
- `sdk/src/runtime/client/navigation.ts` — adds data-reload and shouldIntercept
- `sdk/src/runtime/client/navigation.test.ts` — unit tests
- `sdk/src/runtime/client/navigationCache.ts` — skips prefetch for full-reload links
- `sdk/src/runtime/client/navigationCache.test.ts`— unit tests
- `playground/full-reload-opt-out/` — new example with E2E tests
### Unit tests
```
cd sdk
pnpm test
```
### Playground E2E tests
```
pnpm test:e2e full-reload-opt-out/__tests__/e2e.test.mts
```
### Manual test
```
cd playground/full-reload-opt-out
pnpm dev
```
- Open the home page. The background is light blue.
- Click "Go to Admin (shouldIntercept reload)". The browser does a full reload and the background becomes amber.
- Click "Go to Admin Details (soft nav)". No reload happens; the background stays amber.
- Click "Go to Home (data-reload)". The browser does a full reload and the background becomes blue again.
### Test results
- SDK unit tests: 639 passed, 1 skipped
- Dev E2E tests for the new playground pass
### What we did not do
We dropped the navigate(href, { reload: true }) option. If you want a programmatic full reload from inside your own code, you can already set window.location.href = href directly.


playground/full-reload-opt-out/src/app/pages/AdminPage.tsx
```diff
@@ -0,0 +1,17 @@
+import type { RequestInfo } from "rwsdk/worker";
+
+export function AdminPage({ ctx }: RequestInfo) {
+  ctx;
+  return (
+    <div>
+      <h1>Admin Page</h1>
+      <p>The admin section uses a warm amber stylesheet.</p>
+      <a href="/" id="home-link" data-reload>
```
justinvdm, line 9: Since the `shouldIntercept` we defined will trigger when we exit/leave admin already, we won't know if the test case for the `data-reload` case worked because of `data-reload` logic working (it might be that it works purely because of the `shouldIntercept` logic. Maybe we need to have `data-reload` be a link on home + leading to non admin route, or an admin route + leading to another admin route.

chrsnymn: Thanks for the catch. I will update the test to something like this:
```tsx
<a href="/admin/details" id="admin-details-reload-link" data-reload>
  Go to Admin Details (data-reload, full reload)
</a>
```

PR #1280, opened 2026-08-13
fix(client): recover stale use-client lookup misses

### LINKED ISSUE #1279: onModuleNotFound recovery never fires for a use-client lookup-table miss (only for dynamic-import fetch failures)
`onModuleNotFound` recovery never fires for a use-client **lookup-table miss** — only for dynamic-import fetch failures. Any deploy that adds a `'use client'` component white-screens already-open tabs.

**Version:** rwsdk 1.6.0

## What happens

A browser holding the bundle from deploy *N* receives an RSC payload from deploy *N+1* that references a client component added by that deploy. The loaded bundle's `useClientLookup` has no entry for the id, so `loadModule` throws:

```
(client) No module found for '/src/shared/components/BannerStack.tsx' in module lookup for "use client" directive
```

React surfaces it as an uncaught render error → white screen. With `initClient({ onModuleNotFound: 'reloadWhenReady' })` configured, nothing reloads.

## Why

`dist/runtime/imports/client.js`:

```js
export const loadModule = memoizeOnId(async (id) => {
    const moduleFn = useClientLookup[id];
    if (!moduleFn) {
        throw new Error(`(client) No module found for '${id}' ...`);  // ← outside the try
    }
    try {
        return await moduleFn();
    }
    catch (error) {
        if (isDynamicImportFailure(error) && isRecoveryConfigured()) {
            startRecovery("module-not-found");   // ← the only call site in the package
            return new Promise(() => { });
        }
        throw error;
    }
});
```

The lookup miss throws *before* the `try`, so it can never reach `startRecovery`. And the one gate that does reach it, `isDynamicImportFailure` (`dist/runtime/client/recovery.js:187`), requires a **`TypeError`** whose message contains `"dynamically imported module"`:

```js
export function isDynamicImportFailure(error) {
    return (error instanceof TypeError &&
        typeof error.message === "string" &&
        error.message.includes("dynamically imported module"));
}
```

The lookup-miss error is a plain `Error` with a different message, so it fails that test on both clauses.

Net effect: recovery covers *"the chunk 404'd"* but not *"the id isn't in the table"* — even though the reason string passed to `startRecovery` is literally `"module-not-found"`, which is the case it doesn't handle.

The two are distinct stale-client failure modes:

| | trigger | covered today |
|---|---|---|
| stale HTML → new bundle | hashed chunk URL 404s | ✅ `TypeError: Failed to fetch dynamically imported module` |
| stale bundle → new RSC payload | deploy added a `'use client'` file | ❌ lookup miss, throws past the recovery branch |

The second fires on *any* deploy that adds a client component, which makes it the more routine of the two.

## Suggested fix

Route the miss through the same recovery path:

```js
export const loadModule = memoizeOnId(async (id) => {
    const moduleFn = useClientLookup[id];
    if (!moduleFn) {
        if (isRecoveryConfigured()) {
            startRecovery("module-not-found");
            // Stall so React doesn't crash before the reload.
            return new Promise(() => { });
        }
        throw new Error(`(client) No module found for '${id}' in module lookup for "use client" directive`);
    }
    try {
        return await moduleFn();
    } catch (error) {
        if (isDynamicImportFailure(error) && isRecoveryConfigured()) {
            startRecovery("module-not-found");
            return new Promise(() => { });
        }
        throw error;
    }
});
```

`reloadWhenReady`'s poll-until-ready loop is the right behaviour here: the payload naming the missing id came from the new deploy, so the route is generally ready immediately, and the poll costs nothing when it is.

Happy to open a PR if the shape looks right.

## Workaround

For anyone hitting this before a release lands — match the message in your own `hydrateRootOptions.onUncaughtError` and reload once, guarded by `sessionStorage` so a permanently stale bundle can't loop:

```ts
const MODULE_LOOKUP_MISS = /No module found for .+ in module lookup/i
```

Note this also needs handling in whatever environments you've switched fully onto SDK recovery — if you disabled your own mitigations there (as we had), that's precisely where you have no coverage at all.

chrsnymn:
## Problem

After a deployment adds a new `"use client"` component, a browser tab still running the previous build can encounter a white screen during client-side navigation.

The new deployment's RSC response refers to the new component, but the stale browser tab's in-memory `useClientLookup` does not contain its module ID. `loadModule()` currently throws immediately:

```text
(client) No module found for '<id>' in module lookup for "use client" directive
```

This happens before the existing dynamic-import recovery path. As a result, `onModuleNotFound: "reloadWhenReady"` is never triggered, the error reaches React, and the page becomes blank until manually reloaded.

## Solution

Route missing lookup entries through the existing client recovery flow when `onModuleNotFound` is configured.

- Extract the module-loading decision into `loadClientModule()`.
- If the requested ID is absent from `useClientLookup`, start recovery instead of throwing into React.
- Reuse `suspendForRecovery()` for both lookup misses and failed dynamic imports.
- Keep the module load pending while the recovery controller checks the current route and reloads the document.
- Preserve the existing lookup error when recovery is not configured.
- Continue rethrowing unrelated module errors so application bugs do not become reload attempts.

The pending promise does not block the browser. Once recovery reloads the page, the old document, React root, module state, and pending promise are discarded together.

### Unit coverage

Added tests for:

- preserving the lookup error when recovery is not configured;
- suspending a lookup miss when recovery is configured;
- suspending a failed dynamic import;
- rethrowing unrelated module errors.

```sh
CI=true pnpm --dir sdk exec vitest run \
  --config vitest.config.mts \
  src/runtime/imports/loadClientModule.test.ts
```

Result: 4 tests passed.

### Production redeploy E2E

Extended the existing `module-not-found-recovery` playground test to cover both stale-deployment failure modes:

1. The stale lookup points to an old client chunk that no longer exists.
2. A new deployment adds a client component whose ID is absent from the stale lookup.

The second case is reproduced naturally across production builds. The test confirms that:

- recovery performs a full document reload;
- state belonging to the stale document is discarded;
- the lookup error does not reach React as an uncaught error;
- the new component renders and remains interactive after recovery.

```sh
CI=true GITHUB_EVENT_NAME=pull_request RWSDK_SEQUENTIAL=1 \
  pnpm test:e2e module-not-found-recovery
```

Result: 1 test passed.

### Additional verification

```sh
CI=true pnpm --filter module-not-found-recovery types
```

The SDK production build also passes as part of the E2E command.

## Documentation

Updated the client-side recovery architecture document to cover:

- stale lookup entries;
- why the module promise intentionally remains pending;
- how the reload disposes of the old promise;
- visible pending-navigation behavior;
- the remaining reload-loop limitation for permanently broken deployments.

Fixes #1279

### DIFF
--- a/docs/architecture/clientSideRecovery.md
+++ b/docs/architecture/clientSideRecovery.md
@@ -49,9 +49,9 @@ Applications that want to reload on disconnect can do so through `onStatusChange
 
 ## Hooking the failure path
 
-Every `"use client"` module is loaded through the framework's module lookup in `sdk/src/runtime/imports/client.ts`. `loadModule()` starts recovery when the requested ID is absent from that lookup. It also starts recovery when invoking a known lookup entry fails with a dynamic import error whose message matches `dynamically imported module`. Other errors are rethrown so application failures do not turn into reload attempts.
+Every `"use client"` module is resolved through the framework's module lookup in `sdk/src/runtime/imports/client.ts`, which delegates loading to `loadClientModule()` in `sdk/src/runtime/imports/loadClientModule.ts`. The loader starts recovery when the requested ID is absent from the lookup. It also starts recovery when a known lookup entry fails with a dynamic import error whose message matches `dynamically imported module`. Other errors are rethrown so application failures do not turn into reload attempts.
 
-Both recovery paths call `suspendForRecovery()`. It returns a promise that intentionally remains pending, which keeps React's Suspense boundary suspended while the controller waits to reload. It does not block the browser's main thread. When recovery reloads the page, the browser replaces the entire document and JavaScript environment, including the old React root and the pending promise.
+Both recovery paths call `suspendForRecovery()`. It returns a promise that intentionally remains pending, which keeps React waiting for the module load while the controller prepares to reload. It does not block the browser's main thread. When recovery reloads the page, the browser replaces the entire document and JavaScript environment, including the old React root and the pending promise.
 
 While recovery is waiting, the previous page can remain visible even though the address bar contains the destination URL, and navigation indicators can remain pending. During initial hydration, server-rendered HTML can remain visible without becoming interactive. If a deployment is permanently broken, a fresh document can encounter the same failure and start recovery again; this flow does not persist a reload-loop counter across documents.
 
--- a/docs/architecture/clientSideRecovery.md
+++ b/docs/architecture/clientSideRecovery.md
@@ -6,6 +6,8 @@ A RedwoodSDK application is split between a worker that runs on the server and a
 
 The most common way this surfaces is after a redeploy. RedwoodSDK ships client components as individual JavaScript files whose filenames include content hashes. A new build renames those files. A tab that was opened before the redeploy still references the old names, so when it later tries to load a chunk it has not yet needed, the browser fails to fetch it and React crashes into a blank page.
 
+A stale tab can also receive new page data during client-side navigation. If the new deployment added a client component, that response can refer to a module ID that is absent from the old tab's in-memory lookup. No JavaScript request is made in this case because the old client does not know which file contains the new module. Without recovery, the lookup error reaches React and produces the same blank page.
+
 This can also happen outside of deploys. A CDN or edge node can serve a stale HTML shell that references chunks the origin no longer has. A long-lived tab can sit idle long enough that its assets are evicted from the edge. In each case the tab is holding code that may no longer be valid, and the safest recovery is to reload once the application is confirmed reachable.
 
 ## Why immediate reload is not enough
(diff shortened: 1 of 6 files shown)


docs/architecture/clientSideRecovery.md
```diff
@@ -47,7 +49,11 @@ Applications that want to reload on disconnect can do so through `onStatusChange
...
+
+Both recovery paths call `suspendForRecovery()`. It returns a promise that intentionally remains pending, which keeps React waiting for the module load while the controller prepares to reload. It does not block the browser's main thread. When recovery reloads the page, the browser replaces the entire document and JavaScript environment, including the old React root and the pending promise.
+
+While recovery is waiting, the previous page can remain visible even though the address bar contains the destination URL, and navigation indicators can remain pending. During initial hydration, server-rendered HTML can remain visible without becoming interactive. If a deployment is permanently broken, a fresh document can encounter the same failure and start recovery again; this flow does not persist a reload-loop counter across documents.
```
justinvdm, line 56: Can we make keep this doc more high-level arch? We shouldn't need to refer to programming symbols or file paths, we're talking about the architecture narrative and narrative.

So in this case, great to mention we do recovery both if its not in the lookup (and when that case happens), as well as for abritrary loading errors (and examples of when that happens). But the exact function names, code file paths, etc for the implementation, are TMI.

justinvdm:
Minor docs comment, otherwise looks awesome!

docs/architecture/clientSideRecovery.md
chrsnymn: Cool will update! ta

justinvdm:
Perfect! 🍰

PR #1286, opened 2026-08-26
fix(runtime): preserve UTF-8 characters in streamed HTML

### LINKED ISSUE #1283: Multibyte characters corrupted in streamed SSR HTML (decoder state dropped when switching to raw pass-through)
## Summary

Multibyte UTF-8 characters in server-rendered HTML can be deterministically corrupted
(e.g. `•` U+2022 renders as `��`). It happens in both `vite dev` and production builds.

**Affected versions:** 1.7.0 and 1.7.2 (both verified).

## Root cause

`dist/runtime/lib/stitchDocumentAndAppStreams.js`, function `splitStreamOnFirstNonHoistedTag`:

Chunks from the app stream are decoded with a streaming `TextDecoder`
(`decoder.decode(value, { stream: true })`) while scanning for the first
non-hoisted tag. Once the tag is found, the **remainder of the stream is
forwarded as raw bytes**:

```js
while (true) {
    const { done, value } = await sourceReader.read();
    if (done) { appBodyController.close(); return; }
    appBodyController.enqueue(value); // raw bytes from here on
}
```

If the last decoded chunk ends in the middle of a multibyte character, the
`TextDecoder` retains its lead byte(s) internally — and they are **silently
dropped** when the code switches to raw forwarding. The continuation bytes of
that character then go out raw at the start of the next chunk, producing
invalid UTF-8 in the response. Observed result for U+2022 (`E2 80 A2`) split as
`E2 | 80 A2`: the lead byte is swallowed, the two orphan continuation bytes
render as `��`.

This is easy to hit in practice: React's `renderToReadableStream` emits a large
first chunk (the whole synchronous shell), so the chunk boundary can fall in
the middle of any multibyte character anywhere in the page (bullets, accented
letters, emoji). The corruption is deterministic for a given page.

## Reproduction

Render any page whose HTML is large enough that the first source chunk ends
mid-multibyte-character (e.g. a table repeating `"••• 1234"` masked phone
numbers). Fetch the page and search the response bytes for `EF BF BD`
(U+FFFD): one character will be corrupted at the first chunk boundary.

## Fix

Keep the remainder of the stream flowing through the same decoder (preserving
its state), re-encode, and flush at the end:

```js
while (true) {
    const { done, value } = await sourceReader.read();
    if (done) {
        const tail = encoder.encode(decoder.decode()); // flush pending bytes
        if (tail.length) appBodyController.enqueue(tail);
        appBodyController.close();
        return;
    }
    appBodyController.enqueue(encoder.encode(decoder.decode(value, { stream: true })));
}
```

We verified this fixes the corruption in both dev and production builds.

## Related hardening (same file)

`stitchDocumentAndAppStreams` shares a **single** `TextDecoder` instance across
three different streams (hoisted / outer document / inner app) as phases
interleave, and its later phases (`inner-suspended`, `outer-end`) also forward
raw bytes after having decoded earlier chunks of the same streams. Both are the
same class of bug (decoder state dropped / polluted across sources). We applied
per-stream decoders and decoded forwarding there too.

We're running these changes via `patch-package` in production. Happy to open a
PR if useful.

[rwsdk+1.7.2.patch](https://github.com/user-attachments/files/31078602/rwsdk%2B1.7.2.patch)

chrsnymn:
Fixes #1283.

## Problem

Server-rendered HTML could corrupt characters such as bullets, accented letters, or emoji when their UTF-8 bytes were divided between stream chunks.

The HTML stitcher started decoding the application stream while looking for metadata and rendering markers. After finding the relevant boundary, it switched to forwarding later chunks directly.

If the decoder was holding the beginning of an incomplete character during that switch, those bytes were left behind. The browser then received only the remaining bytes and displayed replacement characters such as `��`.

This is different from the previously fixed HTML tag-boundary problem. That issue involved complete text divided in the middle of a tag such as `<t` / `itle>`. This issue occurs earlier, when the bytes have not formed a complete character yet.

## Solution

Keep each stream flowing through its own decoder for as long as decoding is required:

- Continue decoding the application stream instead of abandoning pending character bytes.
- Give the document, application body, and hoisted metadata independent decoder state.
- Flush each decoder when its stream ends.
- Continue decoding the final document content instead of switching to unsafe direct forwarding.

The existing byte-stream design, metadata handling, Suspense ordering, and early-hydration behavior remain unchanged.

## Reproduction

Extended the existing metadata-hoisting playground with a realistic table containing values such as:
```
Customer 42: ••• 1234
```

React naturally places a stream boundary inside one of the bullet characters in development.
Before the fix, the response contained a Unicode replacement character. After the fix, all characters remain intact.
### Focused stream tests
```
cd sdk
pnpm exec vitest --run src/runtime/lib/stitchDocumentAndAppStreams.test.ts
```
Result: 25 tests passed.

### Full SDK test suite
```
cd sdk
pnpm test
```
Result: 620 tests passed, 1 skipped.

### SDK build
```
cd sdk
pnpm build
```
Result: passed.

### Development E2E
```
RWSDK_SKIP_DEPLOY=1 pnpm test:e2e playground/meta-hoisting/__tests__/e2e.test.mts
```
Result: 2 tests passed, 1 deployment test skipped.

### Repository checks
```
pnpm check
```
Result: passed.

### Local production validation
Built and served the metadata-hoisting playground using Vite’s local production preview.
The response:
- returned HTTP 200;
- contained intact values for Customer 1 and Customer 100;
- contained no Unicode replacement characters;
- contained no raw EF BF BD replacement-character byte sequences.
No Cloudflare deployment was performed.

## Performance
Local measurements did not reveal an obvious performance regression.
The benchmarks were not designed to establish a statistically significant speed difference, so this PR makes no claim that the fix is faster or slower. Performance is not part of the justification for the change.

## Documentation
Updated the streaming architecture documentation to explain:
- why byte chunks are not necessarily character boundaries;
- why separate streams need separate decoder state;
- why decoders must be flushed when their streams end;
- how the existing Suspense and early-hydration ordering is preserved.


justinvdm (approved):
🍰 !

chrsnymn:
Flakey tests. Last commit code changes was green so merging.

